PR #16, opened 2024-04-08
-- linux: add and use linker version script

VDanielEdwards:
## Subject

Adds a version script to the linkage of the main SIL Kit library which only makes
- the C symbols (i.e., `SilKit_*`)
- and the 'legacy' C++ symbols (all functions defined in `SilKit/source/LegacyAbi.cpp`)
visible in the `.so`.

This works for both, the Release and Debug builds.

The resulting library was tested with the demos from this repository, the adapters, and the simulator.

- [ ] Changelog.md updated

### DIFF
--- a/SilKit/source/silkit.map
+++ b/SilKit/source/silkit.map
@@ -1,0 +1,38 @@
+/*
+ * SPDX-FileCopyrightText: 2024 Vector Informatik GmbH
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+
+/* This version script does not apply symbol versioning, it only changes symbol visibility! */
+{
+    global:
+        SilKit_*;
+
+        extern "C++" {
+            "SilKit::Config::ParticipantConfigurationFromString(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)";
+            "SilKit::Config::ParticipantConfigurationFromFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)";
+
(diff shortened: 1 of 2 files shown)


SilKit/source/silkit.map
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+
+/* This version script does not apply symbol versioning, it only changes symbol visibility! */
+{
+    global:
+        SilKit_*;
+
+        extern "C++" {
+            "SilKit::Config::ParticipantConfigurationFromString(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)";
```
MariusBgm, line 14: Thanks, looks better now, with the  demangled / human readable declarations. 
I wonder if we have to explicitly unalias the `std::string`, I guess its some binutils limitation?

VDanielEdwards: Since `ld` only knowns about symbols and doesn't understand C or C++ headers, it requires all parts relevant for the mangled name, which - unfortunately - is the full template with all relevant template arguments and the actual namespace where the template lives (the ABI namespace `std::__cxx11`).

@MariusBgm Do we support compiling with `libc++` instead of `libstdc++` as the standard library? If yes, then we might need two entries per C++ symbol (due to the inline namespace).

MariusBgm: Well, we do support compiling against libc++ from source. 
However, this only affects legacy symbols, so users building from source shouldn't be affected. 
I'm ok with only supporting the legacy symbols for our Ubuntu binary deliveries (e.g. glibc/libstdc++ builds)

MariusBgm (approved):
LGTM!

PR #18, opened 2024-04-08
ITest_Dashboard: delay Stop() in SimStepHandler

VJanKraemer:
## **Subject**
The netsim_coordinated test would immediately call the stop function of the simulation participant which would result in the dashboard 'seeing' the stop state before the running state and the test would fail.

This delays the Stop() until at least on step duration has passed (1 step) which fixes the issue of dashboard missing states.
>

- [ ] Changelog.md updated


VDanielEdwards (approved):
LGTM! 👍

PR #20, opened 2024-04-08
-- mingw: use ifstream constructor taking a const wchar_t *

VDanielEdwards:
## Subject

In the C-API `SilKit_ParticipantConfigurationFromFile` implementation for `_WIN32` platforms we use a non-standard overload of the `std::ifstream` constructor which takes a `std::wstring` (to allow unicode symbols in the path). This PR switches to the overload taking a `const wchar_t *` which is provided on MSVC as well as MinGW.

- [ ] Changelog.md updated


VJanKraemer (approved):
Tested it with mingw64 which build clean and all tests passed 👍

KonradBkd (approved):

PR #24, opened 2024-04-12
Fix flakey ITest_Abort

KonradBkd:
## Subject

The subtest test_Abort_Stopped_Simulation_Sync in ITest_Abort is flakey and possibly fails (locally reproducible on clang-release builds after several iterations).

The test had wrong expectations for the emerging ParticipantStates:
In the test, the system controller stops the simulation and aborts in the stop handler.
However it can happen that a participant is done with the stop before the abort arrives.

This is fixed by extending the expectations for the participants states that can emerge on a valid path from stop->shutdown.


VDanielEdwards (approved):
LGTM! 👍 The new expectation makes sense!

PR #43, opened 2024-05-15
-- utilities: signal monitor: do not call signal handler upon destruction

VDanielEdwards:
## Subject

The `sil-kit-system-controller` currently crashes (SEGFAULT) if the simulation is stopped externally. This happens because the handler is called when the `SignalMonitor` instance is destroyed, and the handler calls the logging functions on a destroyed object.

The fix checks that the signal number is not the 'default' / 'invalid' signal number before calling the handler.

Please run the system controller and a participant that can be stopped 'by hand' without exiting the process. Then, when the participant stops, the system controller should exit normally. Check the exit code (PowerShell: `$LASTEXITCODE`, shell: `$?`).

- [ ] Changelog.md updated

### DIFF
--- a/SilKit/source/util/SignalHandler.cpp
+++ b/SilKit/source/util/SignalHandler.cpp
@@ -45,6 +45,8 @@ BOOL WINAPI systemHandler(DWORD ctrlType);
 
 class SignalMonitor
 {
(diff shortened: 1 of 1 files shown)


VJanKraemer (approved):
LGTM 👍 That was a quick bugfix

PR #52, opened 2024-06-07
SILKIT-1535: Configuration for structured logging

VLukasBecker:
## Subject
 Adds the configuration option to output log messages in a structure JSON format.

- [ ] Changelog.md updated


VDanielEdwards (changes requested):
Great work! A little bit of bikeshedding and some additional work to avoid bugs in the formatting under certain circumstances 😉

SilKit/source/services/orchestration/SystemStateTracker.cpp
```diff
@@ -177,6 +177,11 @@ auto SystemStateTracker::UpdateParticipantStatus(const ParticipantStatus& newPar
     {
         result.participantStateChanged = true;
 
+      /* Log::LoggerMessage lm{_logger, Log::Level::Trace};
```
VDanielEdwards, line 180: Was this a reminder for something? 😸

SilKit/source/core/mock/participant/MockParticipant.hpp
```diff
@@ -427,6 +427,11 @@ class DummyParticipant : public IParticipantInternal
         return "";
     }
 
+    Services::Logging::ILoggerInternal* GetLoggerInternal() override
+    {
+        return nullptr; // todo fix
```
VDanielEdwards, line 432: `MockLogger` still needs to implement `ILoggerInternal`, right?

SilKit/source/core/participant/Participant.hpp
```diff
@@ -300,6 +300,10 @@ class Participant : public IParticipantInternal
 
     std::string GetServiceDescriptorString(
         SilKit::Experimental::NetworkSimulation::ControllerDescriptor controllerDescriptor) override;
+    auto GetLoggerInternal() -> Services::Logging::ILoggerInternal* override
```
VDanielEdwards, line 303: Please put the implementation into `Participant_impl.hpp`

SilKit/source/core/internal/LoggingDatatypesInternal.hpp
```diff
@@ -92,18 +99,54 @@ std::ostream& operator<<(std::ostream& out, const SourceLoc& sourceLoc)
...
 }
 
+
 std::ostream& operator<<(std::ostream& out, const LogMsg& msg)
 {
-    out << "LogMsg{logger=" << msg.logger_name << ", level=" << msg.level
+    out << "LogMsg{logger=" << msg.loggerName << ", level=" << msg.level
         << ", time=" << msg.time.time_since_epoch().count() << ", source=" << msg.source << ", payload=\""
-        << msg.payload << "\""
+        << msg.payload << msg.keyValues << "\""
```
VDanielEdwards, line 149: I think the formatting here isn't correct, the map written between the quotes surrounding the `payload` contents.

SilKit/source/services/logging/Logger.cpp
```diff
@@ -80,10 +277,29 @@ class SilKitRemoteSink
...
     //  this will cause a fairly unintuitive exception in spdlog.
-    _logger = spdlog::create<spdlog::sinks::null_sink_st>(participantName);
+    for (auto sink : _config.sinks)
+    {
+        if (sink.type == Config::Sink::Type::Remote)
+        {
+            _loggerRemote = std::make_shared<RemoteLogger>(sink.level, participantName); 
+        }
+        else
+        {
```
VDanielEdwards, line 290:
```suggestion
        {
            // NB: logger gets dropped from registry immediately after creating so that two participant with the same
            // participantName won't lead to a spdlog exception because a logger with this name does already exist.

```

Move the comment up here and ...

SilKit/source/services/logging/Logger.cpp
```diff
@@ -80,10 +277,29 @@ class SilKitRemoteSink
...
+            if (sink.format == Config::Sink::Format::Json 
+                && nullptr == _loggerJson)
+            {
+                _loggerJson = spdlog::create<spdlog::sinks::null_sink_st>(participantName + "_Json");
```
VDanielEdwards, line 294: ... use only the participant name for the JSON and ...

```suggestion
                _loggerJson = spdlog::create<spdlog::sinks::null_sink_st>(participantName);
                spdlog::drop(participantName);
```

SilKit/source/services/logging/Logger.cpp
```diff
@@ -80,10 +277,29 @@ class SilKitRemoteSink
...
+            }
+            if (sink.format == Config::Sink::Format::Simple 
+                && nullptr == _loggerSimple)
+            {
+                _loggerSimple = spdlog::create<spdlog::sinks::null_sink_st>(participantName + "_Simple");
```
VDanielEdwards, line 299:
```suggestion
                _loggerSimple = spdlog::create<spdlog::sinks::null_sink_st>(participantName);
                spdlog::drop(participantName);
```

SilKit/source/services/logging/Logger.cpp
```diff
@@ -80,10 +277,29 @@ class SilKitRemoteSink
...
+            }
+        }
+    }
 
     // NB: logger gets dropped from registry immediately after creating so that two participant with the same
```
VDanielEdwards, line 304: This comment and the call to `spdlog::drop` can be removed if done for each logger implementation beforehand.

SilKit/source/services/logging/Logger.cpp
```diff
@@ -101,11 +281,26 @@ Logger::Logger(const std::string& participantName, Config::Logging config)
     localtime_r(&timeNow, &tmBuffer);
 #endif
 
+    // Defined JSON pattern for the logger output
+    std::string jsonpattern  {R"({"ts":"%Y-%m-%dT%H:%M:%S.%e%z","log":"%n","lvl":"%l", %v })"};
```
VDanielEdwards, line 285: I think we should increase the timestamp resolution to at least microseconds. Especially trace-level logs are created with sub-millisecond deltas.

`spdlog` should support a `%f` (instead of `%e`). It might be debatable if we need nanosecond resolution.

I could also see a format of `<full-seconds-since-epoch>.<subsecond-remainder>` (with `subsecond` being either `microsecond` or `nanosecond`), but there seems to be no good way to do that in `spdlog` (or C++14 without third-party libraries for that matter) with respect to the UNIX epoch.

Could you ask this in the next daily, there could be different opionions to consider.

SilKit/source/services/logging/Logger.cpp
```diff
@@ -22,26 +22,187 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
 // NB: we do not use the windows color sink, as that will open "CONOUT$" and
 //     we won't be able to trivially capture its output in SilKitLauncher.
 #include "spdlog/sinks/ansicolor_sink.h"
 #include "spdlog/sinks/stdout_sinks.h"
 #include "spdlog/sinks/basic_file_sink.h"
 
 #include "SpdlogTypeConversion.hpp"
+#include "spdlog/pattern_formatter.h"
+#include "spdlog/fmt/ostr.h"  // support for user defined types
+#include "spdlog/cfg/env.h"   // support for loading levels from the environment variable
```
VDanielEdwards, line 46: Do we need that header? Is that actually used (loading log-levels from the environment)?

SilKit/source/services/logging/ILoggerInternal.hpp
```diff
@@ -27,11 +27,111 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
+        : _logger(logger)
+        , _level(msg.level)
+        , _msg(msg.payload)
+        , _keyValues(msg.keyValues)
+    {}
+
+    template <typename... Args>
+    void SetMessage(const char* fmt, const Args&... args)
+    {
+        _msg = fmt::format(fmt, args...);
```
VDanielEdwards, line 80: Use [perfect forwarding](https://en.cppreference.com/w/cpp/utility/forward) for passing around parameter packs.

```suggestion
    void SetMessage(const char* fmt, Args&&... args)
    {
        _msg = fmt::format(fmt, std::forward<Args>(args)...);
```

SilKit/source/services/logging/ILoggerInternal.hpp
```diff
@@ -27,11 +27,111 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
+    }
+
+    ILoggerInternal* GetLogger() const
+    {
+        return _logger;
+    }
+
+    void Dispatch()
+    {
+        if ((_minlevel <= _level))
```
VDanielEdwards, line 120: One too many `(...)`

```suggestion
        if (_minlevel <= _level)
```

SilKit/source/services/logging/ILoggerInternal.hpp
```diff
@@ -27,11 +27,111 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
+    }
+
+    std::unordered_map<std::string, std::string> GetKeyValues() const
+    {
+        return _keyValues;
+    }
+
+    bool HasKeyValues() const
+    {
+        return _keyValues.size() > 0 ? true : false;
```
VDanielEdwards, line 105:
```suggestion
        return !_keyValues.empty();
```

SilKit/source/services/logging/ILoggerInternal.hpp
```diff
@@ -27,11 +27,111 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
+    }
+
+    void SetMessage(std::string newMsg)
+    {
+        _msg = newMsg;
+    }
+
+    void AddKeyValue(std::string key, std::string value)
+    {       
+        _keyValues[key] = value;
```
VDanielEdwards, line 90:
```suggestion
        _keyValues[key] = std::move(value);
```

SilKit/source/services/logging/ILoggerInternal.hpp
```diff
@@ -27,11 +27,111 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
+
+    template <typename... Args>
+    void SetMessage(const char* fmt, const Args&... args)
+    {
+        _msg = fmt::format(fmt, args...);
+    }
+
+    void SetMessage(std::string newMsg)
+    {
+        _msg = newMsg;
```
VDanielEdwards, line 85:
```suggestion
        _msg = std::move(newMsg);
```

SilKit/source/services/logging/ILoggerInternal.hpp
```diff
@@ -27,11 +27,111 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
+    {       
+        _keyValues[key] = value;
+    }
+
+    Level GetLevel() const
+    {
+        return _level;
+    }
+
+    std::unordered_map<std::string, std::string> GetKeyValues() const
```
VDanielEdwards, line 98: We generally use trailing-return-type syntax. Returning a value of `std::unordered_map<...>` will copy the map on every access.

```suggestion
    auto GetKeyValues() const -> const std::unordered_map<std::string, std::string>&
```

SilKit/source/services/logging/ILoggerInternal.hpp
```diff
@@ -27,11 +27,111 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
+    {
+        return _keyValues;
+    }
+
+    bool HasKeyValues() const
+    {
+        return _keyValues.size() > 0 ? true : false;
+    }
+
+    std::string GetMsgString() const
```
VDanielEdwards, line 108: We don't want to copy the message string each time we access it.

```suggestion
    auto GetMsgString() const -> const std::string&
```

SilKit/source/services/logging/ILoggerInternal.hpp
```diff
@@ -27,11 +27,111 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
+    {
+        return _msg;
+    }
+
+    ILoggerInternal* GetLogger() const
```
VDanielEdwards, line 113: Maybe also rename this method to `GetLoggerInternal` since we're not returning a `ILogger*`.

```suggestion
    auto GetLogger() const -> ILoggerInternal*
```

SilKit/source/services/logging/ILoggerInternal.hpp
```diff
@@ -27,11 +27,111 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
+        , _level(msg.level)
+        , _msg(msg.payload)
+        , _keyValues(msg.keyValues)
+    {}
+
+    template <typename... Args>
+    void SetMessage(const char* fmt, const Args&... args)
+    {
+        _msg = fmt::format(fmt, args...);
+    }
```
VDanielEdwards, line 81: Use [perfect forwarding](https://en.cppreference.com/w/cpp/utility/forward) for passing around parameter packs.

```suggestion
    template <typename... Args>
    void SetMessage(fmt::format_string<Args...> fmt, Args&&... args)
    {
        _msg = fmt::format(fmt, std::forward<Args>(args)...);
    }
```

SilKit/source/services/logging/ILoggerInternal.hpp
```diff
@@ -27,11 +27,111 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
+
+    void SetMessage(std::string newMsg)
+    {
+        _msg = newMsg;
+    }
+
+    void AddKeyValue(std::string key, std::string value)
+    {       
+        _keyValues[key] = value;
+    }
```
VDanielEdwards, line 91: Use [perfect forwarding](https://en.cppreference.com/w/cpp/utility/forward) which may avoid copying.

```suggestion
    template <typename Key, typename Value>
    void AddKeyValue(T&& key, Value&& value)
    {       
        _keyValues[std::forward<Key>(key)] = std::forward<Value>(value);
    }
```

SilKit/source/services/logging/ILoggerInternal.hpp
```diff
@@ -27,11 +27,111 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
+    template <typename... Args>
+    void SetMessage(const char* fmt, const Args&... args)
+    {
+        _msg = fmt::format(fmt, args...);
+    }
+
+    void SetMessage(std::string newMsg)
+    {
+        _msg = newMsg;
+    }
```
VDanielEdwards, line 86: Use [perfect forwarding](https://en.cppreference.com/w/cpp/utility/forward) to avoid copying.

```suggestion
    template <typename T>
    void SetMessage(T&& newMsg)
    {
        _msg = std::forward<T>(newMsg);
    }
```

SilKit/source/services/logging/ILoggerInternal.hpp
```diff
@@ -27,11 +27,111 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
+
+    void AddKeyValue(std::string key, std::string value)
+    {       
+        _keyValues[key] = value;
+    }
+
+    Level GetLevel() const
+    {
+        return _level;
+    }
```
VDanielEdwards, line 96: Trailing-return-types ftw 😸 

```suggestion
    auto GetLevel() const -> Level
    {
        return _level;
    }
```

SilKit/source/services/logging/Logger.cpp
```diff
@@ -22,26 +22,187 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
 namespace Logging {
 
+class LoggerMessage;
+
+struct SimpleLogMessage
+{
+    SimpleLogMessage(const LoggerMessage& m)
+        : m(m)
+    {}
+
```
VDanielEdwards, line 60: The constructors are not necessary if you use aggregate-initialization for construction (see below).

```suggestion
```

SilKit/source/services/logging/Logger.cpp
```diff
@@ -22,26 +22,187 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
+struct JsonLogMessage
+{
+    JsonLogMessage(const LoggerMessage& m)
+        : m(m)
+    {}
+
+    JsonLogMessage(LoggerMessage&& m)
+        : m(m)
+    {}
+
```
VDanielEdwards, line 73: The constructors are not necessary if you use aggregate-initialization for construction (see below).

```suggestion
```

SilKit/source/services/logging/Logger.cpp
```diff
@@ -22,26 +22,187 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
...
+    const LoggerMessage& m;
+};
+
+struct JsonString
+{
+    JsonString(const std::string& m)
+        : m(m)
+    {}
+
```
VDanielEdwards, line 82: The constructor is not necessary if you use aggregate-initialization for construction (see below).

```suggestion
```

SilKit/source/services/logging/Logger.cpp
```diff
@@ -118,32 +313,124 @@ Logger::Logger(const std::string& participantName, Config::Logging config)
...
+    const auto now = log_clock::now();
+    if (nullptr != _loggerJson)
+    {
+        JsonLogMessage myJsonMsg(msg);
+        _loggerJson->log(now, spdlog::source_loc{}, to_spdlog(msg.GetLevel()), fmt::format("{}", myJsonMsg));
+    }
+
+    if (nullptr != _loggerSimple)
+    {
+        SimpleLogMessage myMsg(msg);
```
VDanielEdwards, line 382: Aggregate initialization ftw!

```suggestion
        SimpleLogMessage myMsg{msg};
```

SilKit/source/services/logging/Logger.cpp
```diff
@@ -118,32 +313,124 @@ Logger::Logger(const std::string& participantName, Config::Logging config)
...
+        _loggerJson->flush_on(to_spdlog(_config.flushLevel));
+    }
+}
+
+void Logger::Log(const LoggerMessage& msg)
+{
+    const auto now = log_clock::now();
+    if (nullptr != _loggerJson)
+    {
+        JsonLogMessage myJsonMsg(msg);
```
VDanielEdwards, line 376: Aggregate initialization ftw!

```suggestion
        JsonLogMessage myJsonMsg{msg};
```

SilKit/source/services/logging/Logger.cpp
```diff
@@ -118,32 +313,124 @@ Logger::Logger(const std::string& participantName, Config::Logging config)
...
 }
 
+
 void Logger::Log(Level level, const std::string& msg)
 {
-    _logger->log(to_spdlog(level), msg);
+    const auto now = log_clock::now();
+    if (nullptr != _loggerJson)
+    {
+        JsonString myJsonString(msg);
```
VDanielEdwards, line 420: Aggregate initialization ftw!

```suggestion
        JsonString myJsonString{msg};
```

SilKit/source/services/logging/Logger.hpp
```diff
@@ -56,6 +139,10 @@ class Logger : public ILogger
     // Public interface methods
     //
     // ILogger
+    void Log(const LoggerMessage& msg) override;
+
+    void Log(const LogMsg& msg) override;
```
VDanielEdwards, line 144: Please create a new 'comment block' `// ILoggerInternal` (since these methods aren't present in the `ILogger` interface) and put them there.

SilKit/source/services/logging/ILoggerInternal.hpp
VLukasBecker: Is it necessary to have a template type for the Key? String should be sufficient.

SilKit/source/services/logging/Logger.cpp
```diff
@@ -174,53 +462,47 @@ void Logger::Critical(const std::string& msg)
...
-        auto* remoteSink = dynamic_cast<SilKitRemoteSink*>(sink.get());
-        if (remoteSink)
-        {
-            remoteSink->Disable();
-        }
+        _loggerRemote->DisableRemoteLogging();
     }
 }
 
+/*
```
VDanielEdwards, line 482: Please remove the commented code.

SilKit/source/config/YamlConversion.cpp
```diff
@@ -124,7 +125,50 @@ bool Converter::decode(const Node& node, Sink& obj)
...
+    Node node;
+    switch (obj)
+    {
+    case Sink::Format::Simple:
+        node = "Simple";
+        break;
+    case Sink::Format::Json:
+        node = "Json";
+        break;
+    default:
```
VDanielEdwards, line 141: Usually a `default` case in a `switch` statement is a good idea, but in certain circumstances it can be problematic. If the `switch` is supposed to be exhaustive, the `default` case suppresses the compilers ability to emit a warning if the `enum` is extended without adding a new case to the `switch` statement.

The `default` statement does make sense for exhaustive `switch`-es if the acutal `enum` value comes from "user code" (i.e., code not under our control) - because `C++` does not mandate that the underlying integer is any of the `enum` values.

VDanielEdwards: I'm not 100% sure if we have setup our compiler warnings properly, so you can keep it this way if you like 👍 Just thought I'd mention it.

VDanielEdwards (approved):
Code looks good! 👍 I'm running some manual tests now!

SilKit/source/services/logging/Logger.cpp
```diff
@@ -118,32 +314,124 @@ Logger::Logger(const std::string& participantName, Config::Logging config)
...
+    }
+    if (nullptr != _loggerRemote)
+    {
+        _loggerRemote->Log(now, msg);
+    }
 }
 
+void Logger::LogReceivedMsg(const LogMsg& msg)
+{
+    LoggerMessage loggerMsg{this, msg};
```
VDanielEdwards, line 394: For received messages you have to create a `spdlog::details::log_msg` from the `LogMsg`, similar to how the current code on [main](https://github.com/vectorgrp/sil-kit/blob/main/SilKit/source/services/logging/Logger.cpp#L204) does it. You can adapt the [`to_spdlog`](https://github.com/vectorgrp/sil-kit/blob/main/SilKit/source/services/logging/SpdlogTypeConversion.hpp#L114) conversions for that.

VDanielEdwards: For constructing the `log_msg` use the constructor where the first parameter is a `log_clock::time_point` (see [here](https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/details/log_msg.h#L13)).

VDanielEdwards:
@VLukasBecker I'm going to run a few manual tests across a Linux VM and Windows.

VDanielEdwards:
I love it! As an experiment, I changed the 'Updating participant status' message in `SystemStateTracker.cpp` to use the new `LoggerMessage` and the results are great:

```c++
Log::LoggerMessage lm{_logger, Log::Level::Debug};
lm.SetMessage("Updating participant status");
lm.SetKeyValue("participant name", participantName);
lm.SetKeyValue("old participant state", fmt::to_string(oldParticipantState));
lm.SetKeyValue("new participant state", fmt::to_string(newParticipantState));
lm.Dispatch();
```

The 'simple' logger produces the following output:

```
[2024-07-25 13:17:23.558] [EthernetReader] [debug] Updating participant status, participant name: EthernetWriter, new participant state: Shutdown, old participant state: ShuttingDown
```

The JSON logger produces this (pretty-printed by `jq`):

```json
{
  "ts": "1721906243558938",
  "log": "EthernetReader",
  "lvl": "debug",
  "msg": "Updating participant status",
  "kv": {
    "participant name": "EthernetWriter",
    "new participant state": "Shutdown",
    "old participant state": "ShuttingDown"
  }
}
```

SilKit/source/services/logging/Logger.cpp
```diff
@@ -118,32 +314,146 @@ Logger::Logger(const std::string& participantName, Config::Logging config)
...
+            {
+                stdoutSink->set_level(log_level);
+                _loggerSimple->sinks().emplace_back(std::move(stdoutSink));
+            }
             break;
         }
         case Config::Sink::Type::File:
         {
-            //
             auto filename = fmt::format("{}_{:%FT%H-%M-%S}.txt", sink.logName, tmBuffer);
```
VDanielEdwards, line 341: For now the `.txt` extension is fine (imho), but we might want to change that for the `json` formatted files.

SilKit/source/util/StringHelpers.cpp
```diff
@@ -0,0 +1,64 @@
...
+    std::string result;
+    result.reserve(input.size() * 2); // Reserve enough memory for the result
+
+    size_t i = 0;
+    while (i < input.size())
+    {
+        char character = input[i];
+        switch (character)
+        {
+            case '\\':
```
VDanielEdwards, line 40: During testint I noticed that the escaping code is missing the control characters directly supported by JSON.

```suggestion
            case '\b':
            {
                result += "\\";
                result += 'b';
                break;
            }
            case '\t':
            {
                result += "\\";
                result += 't';
                break;
            }
            case '\n':
            {
                result += "\\";
                result += 'n';
                break;
            }
            case '\f':
            {
                result += "\\";
                result += 'f';
                break;
            }
            case '\r':
            {
                result += "\\";
                result += 'r';
                break;
            }
            case '\\':
```

VDanielEdwards: The [JSON spec](https://www.json.org/json-en.html) for reference.

SilKit/source/services/logging/Logger.cpp
VDanielEdwards: Regarding the extension we could use `.ndjson` for 'Newline-Delimited JSON' as proposed [here](https://github.com/ndjson/ndjson-spec) or alternatively `.jsonl` as proposed [here](https://jsonlines.org/). From what I can see, both propose essentially the same thing, and neither are 'official'.

It's probably not a good idea to use `.json` since that would imply a single JSON value.

But yeah, for now `.txt` should be fine, since it can be opened in a text editor anyway.

VLukasBecker: .jsonl looks good, since VS Code recognizes it as JSON.

VDanielEdwards: @VLukasBecker thats a really good point, I didn't even check that!

SilKit/source/util/StringHelpers.hpp
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2024 Vector Informatik GmbH
```
VDanielEdwards, line 1: Hier braucht es noch den include guard:

```suggestion
#pragma once

// Copyright (c) 2024 Vector Informatik GmbH
```

PR #67, opened 2024-06-28
-- ci: Pass x86 to build action for win32 builds

KonradBkd:
For the build-win workflow, the build-arch: x86 was missing for the win32 job. This parameter is passed to https://github.com/ilammy/msvc-dev-cmd to set up the developer console.


VJanKraemer (approved):
No brainer. Thanks for fixing 👍

PR #72, opened 2024-07-08
-- time sync: Couple virtual time to wall clock

KonradBkd:
## Subject
Introduce wall clock coupling of virtual time. Accessible via participant configuration: 
```
Experimental:
  TimeSynchronization:
    AnimationFactor: 1.0
```

## Description
- Added a _WallClockCoupling_ thread similar to the _WatchDog_ thread that manages the time advances when due
- Prevent sending NextSimTask at other places than the _WallClockCoupling_ thread  if the coupling is active
- Improved performance for mixed system w/ wo/ time synchronization by not sending NextSimTask message to participants that don't use a TimeSyncService:
   - The timeSyncService is created by default with a lifecycle service, although it might be unused
   - Now, controllers can be created internally without performing the message subscriptions (preventing `RegisterSilKitService`)
   - For the TimeSyncService, the subscriptions are then performed when the user calls `CreateTimeSyncService`

- Review code changes
- Review changes to docs in `configuration.rst` and `experimental-configuration.rst`
- Try with any demo: Comment out sleeps in SimStepHandler, use adapted participant config with e.g.

```
Experimental:
  TimeSynchronization:
    AnimationFactor: 100.0

```
- Test is not run in the CI, run `FTest_WallClockCoupling.cpp` manually

- [x] Changelog.md updated

### DIFF
--- a/SilKit/IntegrationTests/ITest_Internals_ParticipantModes.cpp
+++ b/SilKit/IntegrationTests/ITest_Internals_ParticipantModes.cpp
@@ -193,8 +193,7 @@ protected:
         }
 
         auto participantInternal = dynamic_cast<SilKit::Core::IParticipantInternal*>(participant.get());
-        Experimental::Services::Orchestration::ISystemController* systemController =
-            participantInternal->GetSystemController();
+        auto systemController = participantInternal->GetSystemController();
 
         auto systemMonitor = participant->CreateSystemMonitor();
         systemMonitor->AddParticipantStatusHandler([&testParticipant, systemController, logger,
@@ -303,8 +302,7 @@ protected:
         }
 
         auto participantInternal = dynamic_cast<SilKit::Core::IParticipantInternal*>(participant.get());
-        Experimental::Services::Orchestration::ISystemController* systemController =
-            participantInternal->GetSystemController();
+        auto systemController = participantInternal->GetSystemController();
         auto systemMonitor = participant->CreateSystemMonitor();
         systemMonitor->AddParticipantStatusHandler([&testParticipant, systemController, logger,
                                                     this](const Services::Orchestration::ParticipantStatus& status) {
@@ -419,8 +417,7 @@ protected:
             SilKit::CreateParticipantImpl(config, systemControllerParticipantName, registryUri);
 
         auto participantInternal = dynamic_cast<SilKit::Core::IParticipantInternal*>(systemControllerParticipant.get());
-        Experimental::Services::Orchestration::ISystemController* systemController =
-            participantInternal->GetSystemController();
+        auto systemController = participantInternal->GetSystemController();
 
         auto* logger = systemControllerParticipant->GetLogger();
 
--- a/SilKit/source/config/ParticipantConfiguration.hpp
+++ b/SilKit/source/config/ParticipantConfiguration.hpp
@@ -263,6 +263,28 @@ struct Middleware
     double connectTimeoutSeconds{5.0};
 };
 
+
+// ================================================================================
+//  TimeSynchronization
+// ================================================================================
+
+//! \brief Structure that contains experimental TimeSynchronization settings
+struct TimeSynchronization
+{
+    double animationFactor{0.0};
+};
+
+
+// ================================================================================
+//  Experimental
+// ================================================================================
+
+//! \brief Structure that contains experimental settings
+struct Experimental
+{
+    TimeSynchronization timeSynchronization;
+};
+
 // ================================================================================
 //  Root
 // ================================================================================
@@ -301,6 +323,7 @@ struct ParticipantConfiguration : public IParticipantConfiguration
     Tracing tracing;
     Extensions extensions;
     Middleware middleware;
+    Experimental experimental;
 };
 
 bool operator==(const CanController& lhs, const CanController& rhs);
@@ -316,6 +339,8 @@ bool operator==(const Tracing& lhs, const Tracing& rhs);
 bool operator==(const Extensions& lhs, const Extensions& rhs);
 bool operator==(const Middleware& lhs, const Middleware& rhs);
 bool operator==(const ParticipantConfiguration& lhs, const ParticipantConfiguration& rhs);
+bool operator==(const TimeSynchronization& lhs, const TimeSynchronization& rhs);
+bool operator==(const Experimental& lhs, const Experimental& rhs);
 
 } // namespace v1
 } // namespace Config
--- a/SilKit/source/config/ParticipantConfiguration.schema.json
+++ b/SilKit/source/config/ParticipantConfiguration.schema.json
@@ -768,6 +768,26 @@
         }
       },
       "additionalProperties": false
+    },
+    "Experimental": {
+      "type": "object",
+      "description": "Experimental configuration",
+      "properties": {
+        "TimeSynchronization": {
+          "type": "object",
+          "description": "Configuration related to time synchronization",
+          "properties": {
+            "AnimationFactor": {
+              "type": "number",
+              "description": "The animation factor describes how fast the simulation is allowed to run, relative to the local wall clock. For example, if the value is 0.1, SIL Kit will try to run the simulation 10 times faster than the wall clock. The default value 0.0 runs the simulation as fast as possible, i.e., the current behavior.",
+              "minimum": 0.0,
+              "default": 0.0
+            }
+          },
+          "additionalProperties": false
+        }
+      },
+      "additionalProperties": false
     }
   },
   "type": "object"
--- a/SilKit/source/core/participant/Participant_impl.hpp
+++ b/SilKit/source/core/participant/Participant_impl.hpp
@@ -158,7 +158,8 @@ void Participant<SilKitConnectionT>::SetupRemoteLogging()
             Config::InternalController config;
             config.name = "LogMsgReceiver";
             config.network = "default";
-            CreateController<Services::Logging::LogMsgReceiver>(config, std::move(supplementalData), true, logger);
+            CreateController<Services::Logging::LogMsgReceiver>(config, std::move(supplementalData), true, true,
+                                                                logger);
         }
 
         auto sinkIter = std::find_if(_participantConfig.logging.sinks.begin(), _participantConfig.logging.sinks.end(),
@@ -174,11 +175,10 @@ void Participant<SilKitConnectionT>::SetupRemoteLogging()
             config.name = "LogMsgSender";
             config.network = "default";
             auto&& logMsgSender =
-                CreateController<Services::Logging::LogMsgSender>(config, std::move(supplementalData), true);
+                CreateController<Services::Logging::LogMsgSender>(config, std::move(supplementalData), true, true);
 
-            logger->RegisterRemoteLogging([logMsgSender](Services::Logging::LogMsg logMsg) {
-                logMsgSender->SendLogMsg(std::move(logMsg));
-            });
+            logger->RegisterRemoteLogging(
+                [logMsgSender](Services::Logging::LogMsg logMsg) { logMsgSender->SendLogMsg(std::move(logMsg)); });
         }
     }
     else
@@ -255,7 +255,7 @@ auto Participant<SilKitConnectionT>::CreateCanController(const std::string& cano
     Core::SupplementalData supplementalData;
     supplementalData[SilKit::Core::Discovery::controllerType] = SilKit::Core::Discovery::controllerTypeCan;
 
-    auto controller = CreateController<Can::CanController>(controllerConfig, std::move(supplementalData), true,
+    auto controller = CreateController<Can::CanController>(controllerConfig, std::move(supplementalData), true, true,
                                                            controllerConfig, &_timeProvider);
 
     controller->RegisterServiceDiscovery();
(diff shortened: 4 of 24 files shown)


SilKit/IntegrationTests/ITest_CatchExceptionsInCallbacks.cpp
VDanielEdwards, line 1: Unnecessary file permission change (makes it executable).

SilKit/IntegrationTests/ITest_Internals_ParticipantModes.cpp
```diff
@@ -193,7 +193,7 @@ class ITest_Internals_ParticipantModes : public testing::Test
         }
 
         auto participantInternal = dynamic_cast<SilKit::Core::IParticipantInternal*>(participant.get());
-        Experimental::Services::Orchestration::ISystemController* systemController =
+        SilKit::Experimental::Services::Orchestration::ISystemController* systemController =
             participantInternal->GetSystemController();
```
VDanielEdwards, line 197:
```suggestion
        auto systemController = participantInternal->GetSystemController();
```

SilKit/IntegrationTests/ITest_Internals_ParticipantModes.cpp
```diff
@@ -303,7 +303,7 @@ class ITest_Internals_ParticipantModes : public testing::Test
...
```
VDanielEdwards, line 307:
```suggestion
        auto systemController = participantInternal->GetSystemController();
```

SilKit/IntegrationTests/ITest_Internals_ParticipantModes.cpp
```diff
@@ -419,7 +419,7 @@ class ITest_Internals_ParticipantModes : public testing::Test
             SilKit::CreateParticipantImpl(config, systemControllerParticipantName, registryUri);
 
         auto participantInternal = dynamic_cast<SilKit::Core::IParticipantInternal*>(systemControllerParticipant.get());
-        Experimental::Services::Orchestration::ISystemController* systemController =
+        SilKit::Experimental::Services::Orchestration::ISystemController* systemController =
             participantInternal->GetSystemController();
```
VDanielEdwards, line 423:
```suggestion
        auto systemController = participantInternal->GetSystemController();
```

SilKit/source/core/participant/Participant_impl.hpp
```diff
@@ -721,8 +722,8 @@ auto Participant<SilKitConnectionT>::CreateTimeSyncService(Orchestration::Lifecy
     config.name = Discovery::controllerTypeTimeSyncService;
     config.network = "default";
     timeSyncService = CreateController<Orchestration::TimeSyncService>(
-        config, std::move(timeSyncSupplementalData), false, &_timeProvider, _participantConfig.healthCheck,
-        lifecycleService);
+        config, std::move(timeSyncSupplementalData), false, false, &_timeProvider, _participantConfig.healthCheck,
+        lifecycleService, _participantConfig.experimental.timeSynchronization.animationFactor);
```
VDanielEdwards, line 726: All services that can be created using the `CreateController` function tempalte get passed a `IParticipantInternal *` as the first argument. The configuration can be retrieved by the `TimeSyncService` constructor from the participant directly.

SilKit/source/services/orchestration/TimeSyncService.cpp
```diff
@@ -232,19 +260,27 @@ struct SynchronizedPolicy : public ITimeSyncPolicy
...
 TimeSyncService::TimeSyncService(Core::IParticipantInternal* participant, ITimeProvider* timeProvider,
-                                 const Config::HealthCheck& healthCheckConfig, LifecycleService* lifecycleService)
+                                 const Config::HealthCheck& healthCheckConfig, LifecycleService* lifecycleService, double animationFactor)
     : _participant{participant}
     , _lifecycleService{lifecycleService}
     , _logger{participant->GetLogger()}
     , _timeProvider{timeProvider}
     , _timeConfiguration{participant->GetLogger()}
     , _watchDog{healthCheckConfig}
+	, _animationFactor{animationFactor}
```
VDanielEdwards, line 276: Somehow this is indented with tabs.

docs/CHANGELOG.rst
```diff
@@ -12,7 +12,9 @@ The format is based on `Keep a Changelog (http://keepachangelog.com/en/1.0.0/) <
 Added
 ~~~~~
 
-- Network Simulation event flow documentation 
+- Synchronize the virtual time with the system clock.
+  Available in a new experimental section within the Participant Configuration.
+- Event flow documentation of for the Network Simulation API.
```
VDanielEdwards, line 17: `of` or `for`? 😸

VDanielEdwards: Also guaranteed merge conflict for someone 😹

docs/configuration/experimental-configuration.rst
```diff
@@ -0,0 +1,50 @@
+.. _sec:cfg-participant-experimental:
+
+===================================================
+Experimental Configuration
+===================================================
+
+.. |ProductName| replace:: SIL Kit
+
+.. |br| raw:: html
```
VDanielEdwards, line 9: This probably breaks every other generator if we'd ever start using them. Consider using the [line block syntax](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#line-blocks).

SilKit/source/services/orchestration/TimeSyncService.cpp
```diff
@@ -546,6 +612,75 @@ bool TimeSyncService::AbortHopOnForCoordinatedParticipants() const
...
+{
+    _wallClockCouplingThreadRunning = true;
+    _stopFuture = _wallClockCouplingThreadStopPromise.get_future();
+
+    _wallClockCouplingThread = std::thread{[&]() {
+
+        SilKit::Util::SetThreadName("SilKit-WallClockCoupling");
+
+        auto startTime = std::chrono::steady_clock::now();
+        const auto duration = _timeConfiguration.NextSimStep().duration;
```
VDanielEdwards, line 632: This value might not be constant across the simulation. If the user sets the simulation step handler mutliple times, they can specify different durations each time.

SilKit/source/services/orchestration/TimeSyncService.cpp
```diff
@@ -546,6 +612,75 @@ bool TimeSyncService::AbortHopOnForCoordinatedParticipants() const
...
+            auto futureStatus = _stopFuture.wait_for(1ms); // Resolution
+
+            if (futureStatus == std::future_status::ready)
+            {
+                // stop was signaled; stopping thread;
+                return;
+            }
+
+            const auto currentRunDuration = std::chrono::steady_clock::now() - startTime;
+            const auto currentVirtualTime = Now();
```
VDanielEdwards, line 646: The value `currentVirtualTime` is never used.

SilKit/source/services/orchestration/TimeSyncService.cpp
```diff
@@ -546,6 +612,75 @@ bool TimeSyncService::AbortHopOnForCoordinatedParticipants() const
...
+auto TimeSyncService::GetCurrentWallClockSyncPoint() const -> std::chrono::nanoseconds
+{
+    Lock lock{_mx};
+    return _currentWallClockSyncPoint;
+}
+
+void TimeSyncService::StartWallClockCouplingThread()
+{
+    _wallClockCouplingThreadRunning = true;
+    _stopFuture = _wallClockCouplingThreadStopPromise.get_future();
```
VDanielEdwards, line 625: It is probably better to not store the future as a member variable, but only on the stack and move it into the lambda.

```c++
auto stopFuture = _wallClockCouplingThreadStopPromise.get_future();

_wallClockCouplingThread = std::thread{[stopFuture = std::move(stopFuture)]() mutable {
    ...
}};
```

SilKit/source/services/orchestration/TimeSyncService.cpp
```diff
@@ -546,6 +612,75 @@ bool TimeSyncService::AbortHopOnForCoordinatedParticipants() const
...
+
+    _wallClockCouplingThread = std::thread{[&]() {
```
VDanielEdwards, line 627: The catch-all captures, particularly by reference, are in almost all cases a bad idea. They, e.g., make it super-easy to accidentially use stack-variables after the caller returned.

SilKit/source/services/orchestration/TimeSyncService.cpp
```diff
@@ -546,6 +612,75 @@ bool TimeSyncService::AbortHopOnForCoordinatedParticipants() const
...
+            if (futureStatus == std::future_status::ready)
+            {
+                // stop was signaled; stopping thread;
+                return;
+            }
+
+            const auto currentRunDuration = std::chrono::steady_clock::now() - startTime;
+            const auto currentVirtualTime = Now();
+
+            if (_isCoupledToWallClock && currentRunDuration / _animationFactor > nextWallClockSyncPoint)
```
VDanielEdwards, line 648: I'm unsure if this works correctly if the execution isn't continuous, e.g., because the program is interrupted by a debugger.

(written before running / testing / playing with the code)

VDanielEdwards (commented):
Apart from the few cosmetic issues, the only real point to clarify is how the coupling thread works, and if the implemented behavior is the one we want.

Since the whole feature is guarded by an experimental flag, we could also change it after release without breaking non-experimental features.

I'll wait with the final approve until we discussed this with the team.

SilKit/source/services/orchestration/TimeSyncService.cpp
KonradBkd: Ok, I directly get them from `_timeConfiguration.NextSimStep().duration`

SilKit/source/services/orchestration/TimeSyncService.cpp
KonradBkd: I don't see a good way of excluding the debugger time. I could try to capture debugger events but I guess this is highly non-trivial. Another way would be to implement a kind of additive clock (mytime += elapsed) that would be  halted while debugging. This would be not very accurate I guess.. I could live with the behavior that there is a "catching-up phase" after debugging.

docs/configuration/experimental-configuration.rst
KonradBkd: I was trying to get a nice line break before the ".. note", no success with line block syntax. However not so importent, I'll remove the |br| and live with the small visual flaw.

SilKit/source/config/ParticipantConfiguration.hpp
```diff
@@ -316,6 +339,7 @@ bool operator==(const Tracing& lhs, const Tracing& rhs);
 bool operator==(const Extensions& lhs, const Extensions& rhs);
 bool operator==(const Middleware& lhs, const Middleware& rhs);
 bool operator==(const ParticipantConfiguration& lhs, const ParticipantConfiguration& rhs);
+bool operator==(const Experimental& lhs, const Experimental& rhs);
```
VDanielEdwards, line 342:
```suggestion
bool operator==(const TimeSynchronization& lhs, const TimeSynchronization& rhs);
bool operator==(const Experimental& lhs, const Experimental& rhs);
```

SilKit/source/config/ParticipantConfiguration.schema.json
```diff
@@ -768,6 +768,26 @@
...
+      "type": "object",
+      "description": "Experimental configuration",
+      "properties": {
+        "TimeSynchronization": {
+          "type": "object",
+          "description": "Configuration related to time synchronization",
+          "properties": {
+            "AnimationFactor": {
+              "type": "number",
+              "description": "Wall clock slow down factor",
```
VDanielEdwards, line 782:
```suggestion
              "description": "The animation factor describes how fast the simulation is allowed to run, relative to the local wall clock. For example, if the value is 0.1 SIL Kit will try to run the simulation 10 times faster than the wall clock. The default value 0.0 runs the simulation as fast as possible, i.e., the current behavior.",
```

SilKit/source/services/orchestration/TimeSyncService.cpp
```diff
@@ -546,6 +612,95 @@ bool TimeSyncService::AbortHopOnForCoordinatedParticipants() const
...
+
+        while (_wallClockCouplingThreadRunning)
+        {
+            // Wait until the next SimTask is due according to the (animated) wall clock
+            auto currentRunDuration = std::chrono::steady_clock::now() - startTime;
+            auto durationUntilNextWallClockSyncPoint = nextAnimatedWallClockSyncPoint - currentRunDuration;
+            HybridWait(std::chrono::duration_cast<std::chrono::nanoseconds>(durationUntilNextWallClockSyncPoint));
+
+            {
+                Lock lock{_mx};
```
VDanielEdwards, line 673: If I'm not mistaken, the mutex `_mx` is only used to protect access to the `_currentWalLClockSyncPoint`. A `std::atomic<std::chrono::nanoseconds::rep> _currentWalLClockSyncPointNs` would do the job too.

SilKit/source/config/ParticipantConfiguration.schema.json
VDanielEdwards: This is just a proposal. I'm not sure if the `Wall clock slow down factor` is worse than this.

SilKit/source/services/orchestration/TimeSyncService.cpp
KonradBkd: It's also for the `_timeConfiguration.NextSimStep()` which might change at any time.

VDanielEdwards (approved):
Looks good 👍 I'm still a bit torn if the 'as-fast-as-possible' catch-up is the correct thing to do, but since it was discussed with other developers and stakeholders, this is ready to merge imho.

PR #73, opened 2024-07-08
-- ci: build linux releases with GCC on Ubuntu 22.04

MariusBgm:
The Ubuntu 18.04 doesn't work well with the current upload-artifact action. Upgrade to 22.04 for now, until we refactor our builds to only run the CMake step in the target container.

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -39,7 +39,7 @@ jobs:
     environment: public-github-runners
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/mariusbgm/sil-kit-ci-ubuntu-18.04:main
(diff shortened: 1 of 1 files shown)


.github/workflows/build-linux.yml
```diff
@@ -34,8 +34,8 @@ jobs:
           do-package: ${{ inputs.do_package }}
           retention-days: ${{ inputs.retention_days }}
 
-  gcc8-release:
-    name: GCC 8 release Builds for Ubuntu 18.04
+  gcc-release:
+    name: GCC release Builds for Ubuntu 22.04
     environment: public-github-runners
     runs-on: ubuntu-latest
     container:
```
VDanielEdwards, line 41: The image is currently still `18.04`

PR #74, opened 2024-07-08
-- ci: use gcc-release preset for builds

MariusBgm:
the linux-release cmake preset uses a legacy build directory which doesnt work with our CI workflows.


VDanielEdwards:
> 👍 Looks good! We might need to add some exceptions for certain GCC diagnostics though, since GCC 13 (default under Ubuntu 24.04) produces additional warnings in `oatpp` and `fmt`.

The diagnostic in question is `-Wdangling-reference`.

VDanielEdwards (approved):
LGTM! 👍

PR #77, opened 2024-07-18
fixup: Shorten thread name

KonradBkd:
Thread name should be 16 characters or less.


VJanKraemer (approved):
Perfect :) Let's wait for CI confirmation, then we can merge it 👍

PR #82, opened 2024-07-22
ci: fix docs build on jenkins

MariusBgm:
This re-enables windows documentation builds. 
The windows jenkins nodes are now fixed at python 3.9, for running sphinx/ docs target

### DIFF
--- a/SilKit/ci/Jenkinsfile
+++ b/SilKit/ci/Jenkinsfile
@@ -30,6 +30,7 @@ def buildConfigs = [
         Arch:"x64",
         MsvcVersion:"14.1",
         PublishArtifacts: true,
+        BuildDocs: true,
         CmakePreset: "vs141-x64-release",
     ]
     ,[
@@ -38,6 +39,7 @@ def buildConfigs = [
         Arch:"x64_x86",
         MsvcVersion:"14.1",
         PublishArtifacts: true,
+        BuildDocs: true,
         CmakePreset: "vs141-x86-release",
     ]
 
@@ -339,6 +341,7 @@ def doBuild(Map config) {
         def buildTestDebug = config.getOrDefault("TestDebug", false) 
         def triggerAbiCheck = config.getOrDefault("TriggerAbiCheck", false)
         def checkLicenses = config.getOrDefault("checkLicenses", false)
+        def pipenvExtraArgs = ""
 
         def buildEnv = []
         def scmVars = [:]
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -33,6 +33,9 @@ option(SILKIT_BUILD_TESTS "Enable unit and integration tests for the SIL Kit" ON
 option(SILKIT_BUILD_UTILITIES "Build the SIL Kit utility tools" ON)
 option(SILKIT_BUILD_DOCS "Build documentation for the SIL Kit (requires Doxygen and Sphinx)" OFF)
 option(SILKIT_INSTALL_SOURCE "Install and package the source tree" OFF)
+if(SILKIT_INSTALL_SOURCE)
```
KonradBkd, line 36: Jenkins sets SILKIT_INSTALL_SOURCE and SILKIT_BUILD_DOCS based on the buildConfig. I'd suggest we should not link this to be able to create packages without docs.

KonradBkd (changes requested):
@VDanielEdwards added some improvements for the Jenkins build that made sense and got reverted here.

SilKit/ci/Jenkinsfile
```diff
@@ -136,8 +138,8 @@ def isFullBuildRequired(branchName) {
     if(branchName == 'main-mirror' || branchName.startsWith("PR-")) {
         return true;
     }
-    // A user requests a fullbuild by adding a "-full" suffix
-    if(branchName.contains("-full")) {
```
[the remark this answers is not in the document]
VDanielEdwards: If we use `/full` as the suffix then it is impossible to add the suffix to the existing branch name:

```
$ git switch -c dev_something/full
fatal: cannot lock ref 'refs/heads/dev_something/full': 'refs/heads/dev_something' exists; cannot create 'refs/heads/dev_something/full'
```

CMakeLists.txt
VDanielEdwards: Having `SILKIT_INSTALL_SOURCE=ON` and `SILKIT_BUILD_DOCS=OFF` does not lead to an error during configuration, build, or packaging.

If the `if(...)` is kept, we always force users to have python/sphinx/... setup correctly, even if they are not interested in building the documentation at all (but would like to get a properly installed `SilKit.dll` or `libSilKit.so`).

SilKit/ci/Jenkinsfile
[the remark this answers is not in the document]
MariusBgm: I reverted to eagerly. will undo the revert and add the BuildDocs:true to the windows Jenkins jobs. Thanks for pointing this out.

CMakeLists.txt
MariusBgm: marking this as resolved, as I dropped the "revert of 84445e10ab20d9aaf9ea55147c50c2fed68bff3f" from the patch

MariusBgm:
OK, I dropped the whole revert-of-a-previous-commit from this branch. It now concentrates on pinning the python version to 3.9 on windows jenkins hosts.

KonradBkd (approved):
LGTM

PR #93, opened 2024-08-02
SILKIT-1572 core: message aggregation

VJonasHolley:
## Subject

Implementation of a message aggregation mechanism for the VAsioPeer (time sync case only).

## Description

For simulations with time synchronization, user data messages (PubSub, RPC, CAN, Ethernet) are aggregated in the VAsioPeer and send en bloc at the end of the corresponding time step. Messages are only aggregated, once the state of the lifecycle is ReadyToRun.
The aggregation feature can be enabled/disabled via the flag "EnableMessageAggregation" in the participant config file.

NB: The Test_MetricsManager has been deactivated (failure in previous commit). We need to fix this and reactivate the test.

- [ ] Changelog.md updated

### DIFF
--- a/SilKit/source/config/Configuration.hpp
+++ b/SilKit/source/config/Configuration.hpp
@@ -23,16 +23,19 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. */
 
 #include <algorithm>
 #include <iostream>
+#include <sstream>
 #include <stdexcept>
 #include <string>
 #include <tuple>
 #include <vector>
+#include <cctype>
 
 #include "silkit/participant/exception.hpp"
 #include "silkit/SilKitMacros.hpp"
 #include "silkit/services/logging/LoggingDatatypes.hpp"
 
 #include "Optional.hpp"
+#include "StringHelpers.hpp"
 
 namespace SilKit {
 namespace Config {
@@ -57,6 +60,21 @@ enum class NetworkType
 inline auto to_string(NetworkType networkType) -> std::string;
 
 // ================================================================================
+//  Message aggregation declarations
+// ================================================================================
+
+enum class Aggregation : uint32_t
+{
+    Off = 0, // disable aggregation
+    On = 1, // enable aggregation for time synchronization (synchronous and asynchronous case)
+    Auto = 2 // enable aggregation for time synchronization (synchronous case)
+};
+
+inline std::string to_string(const Aggregation& aggregation);
+inline Aggregation from_string(const std::string& aggregationStr);
+inline std::ostream& operator<<(std::ostream& out, const Aggregation& aggregation);
+
+// ================================================================================
 //  Logging service
 // ================================================================================
 
@@ -70,7 +88,6 @@ struct Sink
     };
 
 
-
     enum class Format : uint8_t
     {
         Simple,
@@ -78,8 +95,8 @@ struct Sink
     };
 
     Format format{Format::Simple};
-    Type type{ Type::Remote };
-    Services::Logging::Level level{ Services::Logging::Level::Info };
+    Type type{Type::Remote};
+    Services::Logging::Level level{Services::Logging::Level::Info};
     std::string logName;
 };
 
@@ -225,10 +242,7 @@ auto to_string(NetworkType networkType) -> std::string
 
 bool operator==(const Sink& lhs, const Sink& rhs)
 {
-    return lhs.type == rhs.type
-        && lhs.level == rhs.level
-        && lhs.format == rhs.format
-        && lhs.logName == rhs.logName;
+    return lhs.type == rhs.type && lhs.level == rhs.level && lhs.format == rhs.format && lhs.logName == rhs.logName;
 }
 
 bool operator<(const Sink& lhs, const Sink& rhs)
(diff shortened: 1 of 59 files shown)


SilKit/source/core/vasio/VAsioDatatypes.hpp
```diff
@@ -147,6 +147,13 @@ struct ProxyMessage
     std::vector<uint8_t> payload;
 };
 
+enum class MessageAggregationKind : uint8_t
+{
+    UserDataMessage = 0,
+    TimeSyncMessage = 1,
```
KonradBkd, line 153: I think we need more types that trigger the flush, so we should call this `FlushAggregationMessage` or similar

KonradBkd (changes requested):
Only minor stuff in the code 👍  But I found a conceptual problem with the flush mechanism that happens only on `NextSimTask` (and the size limit). If you have a `lifecycleService->Stop()` or `Pause` in the SimStep, we explicitly prevent sending out the `NextSimTask` to ensure that no SimTask is triggered after these commands (see https://github.com/vectorgrp/sil-kit/blob/main/SilKit/source/services/orchestration/TimeSyncService.cpp#L114). This leads to the behavior that aggregated messages before a Stop in the SimTask are never transmitted. I tested this in the PubSub Demo by adding a Stop in the SimTask. With active aggregation, the Subscriber never receives something. Without aggregation, he does. We should add the `ParticipantStatus` msg to the kinds that trigger the flush, so we ensure that msgs are sent on any participant state change.

What's still completely missing in the PR are tests. For this powerful feature that intervenes deeply in the VAsio code, we need integration tests.

And docs. You might get away with keeping it undocumented for now as the whole thing is experimental :)

SilKit/source/config/Configuration.hpp
```diff
@@ -269,6 +286,50 @@ bool operator==(const MdfChannel& lhs, const MdfChannel& rhs)
...
+        break;
+    default:
+        outStream << "Invalid Aggregation";
+    }
+    return outStream;
+}
+
+inline Aggregation from_string(const std::string& aggregationStr)
+{
+    auto lowerCase = [](auto s) {
```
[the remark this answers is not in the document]
VJonasHolley: Got myself a little bonus ;)

SilKit/source/core/internal/IParticipantInternal.hpp
```diff
@@ -333,6 +333,7 @@ class IParticipantInternal : public IParticipant
                                                                           const std::string& msgTypeName) = 0;
 
     virtual void NotifyShutdown() = 0;
+    virtual void EnableAggregation(bool isBlocking) = 0;
```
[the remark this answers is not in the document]
VJonasHolley: Done.

SilKit/source/core/vasio/AggregationMessageTraits.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include "VAsioDatatypes.hpp"
+
+#include "OrchestrationDatatypes.hpp"
+
+#include "WireCanMessages.hpp"
+#include "WireDataMessages.hpp"
+#include "WireEthernetMessages.hpp"
+#include "WireFlexrayMessages.hpp"
+#include "WireLinMessages.hpp"
```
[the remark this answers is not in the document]
VJonasHolley: Done.

SilKit/source/core/vasio/VAsioDatatypes.hpp
VJonasHolley: Done.

VJonasHolley:
I added messages of type 'ParticipantStatus' in the classification of FlushAggregationMessages (AggregationMessageTraits.hpp). Using the modified PubSubDemo shows that this change solves the 'no-message-reception-problem'.

SilKit/source/core/vasio/VAsioPeer.cpp
```diff
@@ -109,20 +109,60 @@ auto VAsioPeer::GetSimulationName() const -> const std::string&
...
+
+void VAsioPeer::SendSilKitMsgInternal(const std::vector<uint8_t>& blob)
 {
     // Prevent sending when shutting down
     if (!_isShuttingDown && _socket != nullptr)
     {
         std::unique_lock<std::mutex> lock{_sendingQueueMutex};
 
-        _sendingQueue.push_back(buffer.ReleaseStorage());
+        _sendingQueue.push_back(blob);
```
MariusBgm, line 137: This makes a copy because blob is a `const &`. blob should be passed as value and moved. or as an R-value reference.

SilKit/source/config/ParticipantConfiguration_Full.json
```diff
@@ -232,7 +232,8 @@
   },
   "Experimental": {
     "TimeSynchronization": {
-      "AnimationFactor": 1.5
+      "AnimationFactor": 1.5,
+      "EnableMessageAggregation": "Auto"
```
MariusBgm, line 236:
```suggestion
      "EnableMessageAggregation": "Off" 
```
The optimization should be enabled by explicit user request, e.g. when the workload contains a lot of small messages

MariusBgm (changes requested):
Mini review. I will look into it more closely.

SilKit/source/core/vasio/VAsioPeer.hpp
```diff
@@ -130,9 +140,17 @@ class VAsioPeer
...
+    std::vector<uint8_t> _aggregatedMessages;
 
     std::atomic_bool _sending{false};
     Core::ServiceDescriptor _serviceDescriptor;
+
+    bool _useAggregation{false};
+    const size_t _aggregationBufferThreshold{100 * 1000};
+
+    std::unique_ptr<ITimer> _flushTimer;
+    std::chrono::milliseconds _flushTimeout{50};
```
MariusBgm, line 152:
```suggestion
    const std::chrono::milliseconds _flushTimeout{50};
```
that's actually a very important number. Just in case this comes up during debugging simulations runs or from user feedback :-)

SilKit/source/config/ParticipantConfiguration.hpp
```diff
@@ -296,9 +296,9 @@ struct Middleware
 struct TimeSynchronization
 {
     double animationFactor{0.0};
+    Aggregation enableMessageAggregation{Aggregation::Auto};
```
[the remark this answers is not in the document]
VJonasHolley: Thanks for the hint! I changed the default to `Off`.

MariusBgm (approved):
LGTM! thanks for changing the default aggregation to `Off`.

PR #101, opened 2024-08-28
tests: remove MakeTestRegistryUri

MariusBgm:
remove the `MakeTestRegistryUri`  function from all tests.
Use the dynamically allocated registry port instead.

### DIFF
--- a/SilKit/IntegrationTests/SimTestHarness/SimTestHarness.cpp
+++ b/SilKit/IntegrationTests/SimTestHarness/SimTestHarness.cpp
@@ -300,7 +300,7 @@ SimParticipant* SimTestHarness::GetParticipant(const std::string& participantNam
     return _simParticipants[participantName].get();
 }
 
-auto SimTestHarness::GetRegistryUri() -> std::string
+auto SimTestHarness::GetRegistryUri() const -> std::string
 {
     return _registryUri;
 }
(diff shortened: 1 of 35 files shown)


SilKit/IntegrationTests/SimTestHarness/SimTestHarness.hpp
```diff
@@ -132,6 +132,8 @@ class SimTestHarness
     //! \brief Get the SimParticipant by name. If it does not exist yet, create a SimParticipant with the specified name and provide its ParticipantConfiguration as a string.
     SimParticipant* GetParticipant(const std::string& participantName, const std::string& participantConfiguration);
 
+    auto GetRegistryUri() -> std::string;
```
VJonasHolley, line 135:
```suggestion
    auto GetRegistryUri() -> const std::string;
```

SilKit/IntegrationTests/SimTestHarness/SimTestHarness.cpp
```diff
@@ -300,6 +300,11 @@ SimParticipant* SimTestHarness::GetParticipant(const std::string& participantNam
     return _simParticipants[participantName].get();
 }
 
+auto SimTestHarness::GetRegistryUri() -> std::string
```
VJonasHolley, line 303:
```suggestion
auto SimTestHarness::GetRegistryUri() -> const std::string
```

VJonasHolley (commented):
Only 2 minor remarks:

1. Clang Format: The CI shows 6 files, which are not formatted correctly (warnings).
2. For an entire cleanup w.r.t. the MakeTestRegistryUri() method, you could consider removing the (now obsolete) header `include "GetTestPid.hpp"` from the ITests.

PR #103, opened 2024-09-04
cmake: add separate executable for integration tests using internals

VDanielEdwards:
## Subject

Cleanup integration test definitions and split off separate executable for integration tests (and function tests) using internals.

- [ ] Changelog.md updated

### DIFF
--- a/SilKit/IntegrationTests/CMakeLists.txt
+++ b/SilKit/IntegrationTests/CMakeLists.txt
@@ -29,6 +29,7 @@ add_library(ITest_Utils INTERFACE)
 target_include_directories(ITest_Utils
     INTERFACE ${CMAKE_CURRENT_LIST_DIR})
 
+
 target_link_libraries(SilKitIntegrationTests
     PRIVATE SilKit
     PRIVATE S_ITests_STH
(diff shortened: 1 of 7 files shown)


SilKit/IntegrationTests/CMakeLists.txt
```diff
@@ -29,147 +29,112 @@ add_library(ITest_Utils INTERFACE)
...
+)
+
+target_link_libraries(SilKitInternalIntegrationTests
+    PRIVATE S_SilKitImpl
+    PRIVATE S_ITests_STH_Internals
+)
+
+target_link_libraries(SilKitFunctionalTests
+    PRIVATE SilKit
+    PRIVATE S_ITests_STH
```
MariusBgm, line 45: Nice! we're being explicit about the library dependencies upfront.

SilKit/IntegrationTests/Hourglass/CMakeLists.txt
```diff
@@ -46,6 +51,10 @@ endif ()
 
 add_silkit_test_executable(SilKitHourglassTests)
 
+target_compile_definitions(SilKitHourglassTests
+    PRIVATE SIL_KIT_TEST_MAIN_DO_NOT_PRINT_VERSION
```
MariusBgm, line 55: Is this for performance reasons? many small tests and std I/O overhead?

SilKit/cmake/SilKitTest/SilKitTestMain.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include <iostream>
+
+int main(int argc, char** argv)
+{
+    std::cout << "Running main() from SIlKitTestMain.cpp" << std::endl;
+
+    testing::InitGoogleMock(&argc, argv);
+
+#ifndef SIL_KIT_TEST_MAIN_DO_NOT_PRINT_VERSION
+    std::cout << "Running with SIL Kit version '" << SilKit::Version::String() << "' and git hash '"
```
MariusBgm, line 19: could we use a gtest stream instead of std::cout instead? or is this negligable overhead

MariusBgm (approved):
LGTM! left some comments / thoughts

SilKit/IntegrationTests/Hourglass/CMakeLists.txt
VDanielEdwards: No this is because the hourglass tests implement `SilKit_Version_String` (among all the other C-functions) with a call to a `MOCK_METHOD` (see `MockCapi.hpp`, `MockCapi.cpp`, and `MockCapiTest.hpp`.

This code crashes if the underlying global pointer hasn't been initialized (which is only done when the test starts).

SilKit/cmake/SilKitTest/SilKitTestMain.cpp
VDanielEdwards: This is neglegible. In fact the `main` function provided by [`gmock_main`](https://github.com/google/googletest/blob/main/googlemock/src/gmock_main.cc) uses `std::cout` as well ([`gtest_main`](https://github.com/google/googletest/blob/main/googletest/src/gtest_main.cc) uses `std::printf` ❓)

PR #105, opened 2024-09-10
Fix double free reported on Fedora 40

VJanKraemer:
## description
Running the tests under Fedora 40 revealed a double free (confirmed by valgrind) with the initializer_list arg for
	optional_decode_deprecated_alternative
This commit changes the arg to an std::array, which should behave equal to the initializer_list without the syntactic sugar.

### DIFF
--- a/SilKit/source/config/YamlConversion.cpp
+++ b/SilKit/source/config/YamlConversion.cpp
@@ -770,7 +770,9 @@ template <>
 bool Converter::decode(const Node& node, RpcServer& obj)
 {
     obj.name = parse_as<std::string>(node["Name"]);
-    optional_decode_deprecated_alternative(obj.functionName, node, "FunctionName", {"Channel", "RpcChannel"});
+
+    auto deprecatedRPCFields = std::array<std::string, 2> {"Channel", "RpcChannel"};
+    optional_decode_deprecated_alternative(obj.functionName, node, "FunctionName", deprecatedRPCFields);
     optional_decode(obj.useTraceSinks, node, "UseTraceSinks");
     optional_decode(obj.replay, node, "Replay");
     return true;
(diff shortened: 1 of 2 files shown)


SilKit/source/config/YamlConversion.cpp
```diff
@@ -791,7 +793,9 @@ template <>
 bool Converter::decode(const Node& node, RpcClient& obj)
 {
     obj.name = parse_as<std::string>(node["Name"]);
-    optional_decode_deprecated_alternative(obj.functionName, node, "FunctionName", {"Channel", "RpcChannel"});
+
+    auto deprecatedRPCFields = std::array<std::string, 2> {"Channel", "RpcChannel"};
```
VDanielEdwards, line 797:
```suggestion
    auto deprecatedRPCFields = std::array<std::string, 2>{"Channel", "RpcChannel"};
```

SilKit/source/config/YamlConversion.cpp
```diff
@@ -770,7 +770,9 @@ template <>
 bool Converter::decode(const Node& node, RpcServer& obj)
 {
     obj.name = parse_as<std::string>(node["Name"]);
-    optional_decode_deprecated_alternative(obj.functionName, node, "FunctionName", {"Channel", "RpcChannel"});
+
+    auto deprecatedRPCFields = std::array<std::string, 2> {"Channel", "RpcChannel"};
```
VDanielEdwards, line 774:
```suggestion
    auto deprecatedRPCFields = std::array<std::string, 2>{"Channel", "RpcChannel"};
```

VDanielEdwards (approved):

MariusBgm:
Could not be reproduced after Fedora update.

PR #106, opened 2024-09-10
SILKIT-1529 Harmonize CLI of commandline utilities

AndreasRentschler:
## Subject

Commandline utilities Registry, Monitor and System Controller behaved differently for the user:
* Not all used OS signals to shutdown the tool, some required the user to press Enter to shutdown.
* Some utilities by default waited for an additional Enter to terminate after shutdown.

## Description

Changes:
* Commandline utilities Registry, Monitor also use an OS signal handler to detect termination requests.
  Users can now terminate via Ctrl-C instead of Enter key. This behavior was already implemented for the System Controller.
* System Controller option '--non-interactive' is deprecated and enabled by default.
* Registry option '-s, --use-signal-handler' is deprecated and enabled by default if not started as '--windows-service'.
* An interactive mode '-i' can be used to ask user for Enter key to end process after shut down.

Things to check
* Testing: Run demo e.g. Demos/Can, including a Monitor with  `LogFromRemotes: true` and participants logging to it via `Remote` sink.
  * Show help via `-h` of System Controller, Monitor and Registry: Same as on the updated page of our documentation? None of the deprecated options `--non-interactive` and `--signal-handler` listed?
  * Terminate System Controller, Monitor and Registry via `[Ctrl-C]`: Should work
  * Run Monitor and Registry with and without `-i`: When active, is there a prompt on start to press `[Ctrl-C]` to shutdown, and another prompt to press `[Enter]` after shutdown, allowing to read all text? Does `[Enter]` terminate the process?
    (Note: Monitor keeps forwarding remote log even after termination)
  * Try deprecated options: Are `sil-kit-system-controller --non-interactive` and `sil-kit-registry --signal-handler` still accepted but ignored when used?
* Code review

- [X] Changelog.md updated
- [X] Documentation updated (public API changes only)


AndreasRentschler:
As suggested by @VDanielEdwards, an alternative to introduce option `--interactive'  is to deploy a Windows batch script which  pauses our utilities after execution, for example:

`run-and-pause.cmd`
```bat
@echo off

if ["%~1"]==[""] goto usage
%* && pause || pause
exit /b %errorlevel%
goto :eof

:usage
@echo Execute a command line program and then prompt the user to press a key.
@echo Usage: run-and-pause ^<executable^> ^[^<argument1^> ^[^<argument2^> ...^]^]
exit /b 1
```

Downside: When the user uses `[Ctrl+C]` to signal the executed tool to shutdown, the signal is subsequently forwarded to Windows cmd, which asks to `Terminate batch job (Y/N)?`. This behavior cannot be suppressed to my knowledge.

KonradBkd (approved):
Deprication warning in, unwanted flags out, LGTM👍

PR #110, opened 2024-09-13
cmake: simplify presets

MariusBgm:
Remove unused and too complicated presets.
Use CC and CXX env vars to override compiler selections.


MariusBgm:
some issues with the CI / main workflow and the presets were already fixed. will revive this later

MariusBgm:
Resurrecting this. We could really simplify our developer workflows with a more lightweight CmakePresets.json

MariusBgm:
close this in favor of #244

PR #112, opened 2024-09-13
ci: use matrix for sanitizer workflows

MariusBgm:
trying to simplify our workflows as much as possible. Draft


MariusBgm:
will try in my own fork

PR #115, opened 2024-09-18
cmake: make preset build dirs consistent with names

MariusBgm:
ensure the preset name is used for the build directory, e.g. `_build/${presetName}`

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -5,34 +5,20 @@
             "name": "base-env",
             "description": "Basic preset that defines global variables",
             "hidden": true,
-            "generator": "Ninja",
-            "binaryDir": "${sourceDir}/_build/$env{SILKIT_CMAKE_PRESET_NAME}",
-            "installDir": "${sourceDir}/_install/$env{SILKIT_CMAKE_PRESET_NAME}",
-            "environment": {
-                "SILKIT_CMAKE_PRESET_NAME": "$env{SILKIT_CMAKE_PRESET_BUILD_PLATFORM}-$env{SILKIT_CMAKE_PRESET_ARCHITECTURE}-$env{SILKIT_CMAKE_PRESET_BUILD_TYPE}"
-            },
+            "inherits": "mixin-dirs",
             "cacheVariables": { "CMAKE_BUILD_TYPE": "$env{SILKIT_CMAKE_PRESET_BUILD_TYPE}" }
         },
         {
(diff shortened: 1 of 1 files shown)


.github/actions/build-cmake-preset/action.yml
```diff
@@ -107,39 +89,25 @@ runs:
...
-        retention-days: ${{ inputs.retention-days }}
+        retention-days: 2
 
     - name: create cmake package for ${{ inputs.preset-name }}
-      if: ${{ inputs.do-package  == 'true' }}
       id: cmake-package
       run: |
         cmake --build --preset ${{ inputs.preset-name }} --target package
-        echo "package-name=$(basename _build/${{inputs.preset-name}}/_CPack_Packages/*/ZIP/SilKit-*-*/)" >> $GITHUB_OUTPUT
+        echo "package-name=$(basename _build/${{inputs.preset-name}}/_CPack_Packages/*/ZIP/SilKit-*-*/)-${{ inputs.artifact-label }}" >> $GITHUB_OUTPUT
```
VJanKraemer, line 102: Do we need the artifact-label here?
If we check the builds...the toolchain + arch are already in the zip namefile e.g.

- CPack: - package: D:/a/sil-kit/sil-kit/_build/vs141-x86-release/SilKit-4.0.53-Win-x86-VS2017-Release.zip
- CPack: - package: D:/a/sil-kit/sil-kit/_build/vs141-x64-release/SilKit-4.0.53-Win-x86_64-VS2017-Release.zip

Adding the clang-release or  win-32 lable seems reundant? Or am I missing some upload path shenanigans?

VDanielEdwards: I think the tsan/asan/... builds produce artifacts with the same name as the 'normal' builds or am I mistaken? In this case the `name` input to `upload-artifact` should be explicitly different.

MariusBgm: > I think the tsan/asan/... builds produce artifacts with the same name as the 'normal' builds or am I mistaken? In this case the `name` input to `upload-artifact` should be explicitly different.

yes, that's right. I think upgrading to v4 of upload-artifact made `same names` an error, that's why I just unconditionally append the label now. the artifacts are only ever useful when we do debugging, so it shouldn't be a nuisance.

VJanKraemer (approved):

PR #118, opened 2024-09-20
dev: For SimStepAsync, measure execution time until call to CompleteSimStep

KonradBkd:
## Subject

- The Log message "Finished Simulation Step. Execution time was: ..." measures the SimStepHandler execution time. This is misleading when using the  `SimStepHandlerAsync`, because the simulation step is actually not finished.

## Description

- If using the `SimStepHandlerAsync`, the Log Message "Finished Simulation Step. Execution time was: ..." is now printed after the call to `CompleteSimulationStep`.
- The metric to time the SimStepHandler execution has been renamed to _SimStepHandlerExecutionDuration_
- If using the `SimStepHandlerAsync`,  the new metric _SimStepCompletionDuration_ measure the interval between the end of the handler and the call to `CompleteSimulationStep`.

- [x] Changelog.md updated

### DIFF
--- a/SilKit/source/services/orchestration/TimeSyncService.cpp
+++ b/SilKit/source/services/orchestration/TimeSyncService.cpp
@@ -225,7 +225,7 @@ private:
         // If paused, don't request the next sim step. This happens in LifecycleService::Continue()
         if (!_controller.PauseRequested())
         {
-            // With real-time sync, the participant must sigal his readiness for the next step only if the real-time has actually passed.
+            // With real-time sync, the participant must signal his readiness for the next step only if the real-time has actually passed.
             if (!_controller.IsCoupledToWallClock())
             {
                 // The synchronous SimStep API creates the nextSimTask message automatically after the callback.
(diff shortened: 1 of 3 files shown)


VDanielEdwards:
@KonradBkd I think it would be better to add an additional metric that measures the completion duration in contrast to the execution duration. Both measurements are interesting, e.g., async sim. step handler is taking longer than expected due to lock contention.

SilKit/source/services/orchestration/TimeSyncService.cpp
```diff
@@ -499,13 +499,26 @@ void TimeSyncService::ExecuteSimStep(std::chrono::nanoseconds timePoint, std::ch
     _watchDog.Start();
     _simTask(timePoint, duration);
     _watchDog.Reset();
+
+    _simStepCounterMetric->Add(1);
+
+    if (!IsBlocking())
```
VDanielEdwards, line 505: Isn't the `if` condition inverted? I.e., the step immediately completes if the step is sync, since `IsBlocking() == IsSimStepSync()`.

KonradBkd: Yes got this one wrong

KonradBkd:
> @KonradBkd I think it would be better to add an additional metric that measures the completion duration in contrast to the execution duration. Both measurements are interesting, e.g., async sim. step handler is taking longer than expected due to lock contention.

You're right, we should measure both and should be precise with the terminology. I'd say a `LogicalSimStepDuration` spans from the invocation of the handler until 

- SimStepHandler -> handler returns
- SimStepHandlerAsync -> call to `CompleteSimStep`

Whereas, the `SimStepHandlerExecutionDuration` is just around the handler in both modes.

Or, as you suggested, we the split in handler execution time and completion time.

The issue came up when looking at the output by `Trace(_logger, "Finished Simulation Step. Execution time was: {}ms", executionDurationMs.count());`, which is kind of misleading (because too small) when using the SimStepHandlerAsync.

VDanielEdwards (commented):
👍 Great idea!

VDanielEdwards (approved):

PR #119, opened 2024-09-23
ci: don`t run CI and workflows on push to main

MariusBgm:
Make sure that we don't run the `sil-kit-ci` workflow and the individual `build-win` workflows on pushes to main.
Also remove the concurrency group from `build-win`, because it'll get canceled if CI runs in parallel.


VJanKraemer (approved):
LGTM

PR #120, opened 2024-09-25
api: serdes: allow union (de)serialization

VDanielEdwards:
## Subject

This change allows to (de-)serialize unions using the public serdes headers. It does not change the API of the `Serializer` / `Deserializer` classes, but only implements the existing member functions.

The index is 1-based, since a related piece of software already uses 1-based indices for union discriminators.

- [ ] Changelog.md updated

### DIFF
--- a/SilKit/include/silkit/util/serdes/Serializer.hpp
+++ b/SilKit/include/silkit/util/serdes/Serializer.hpp
@@ -38,6 +38,8 @@ inline namespace v1 {
 
 class Serializer
 {
+    using UnionDiscriminator = uint32_t;
+
 public:
     // ----------------------------------------
     // CTOR, DTOR, Copy Operators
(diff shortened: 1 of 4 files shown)


SilKit/include/silkit/util/serdes/Serializer.hpp
```diff
@@ -151,13 +153,26 @@ class Serializer
...
+     * Only the active union member must be serialized prior to calling EndUnion.
+     * \param discriminator 1-based index of the active union member
+     */
+    void BeginUnion(const int discriminator)
     {
-        throw SilKitError("Unions are currently not supported.");
+#ifndef NDEBUG
+        if (discriminator < 0)
+            throw SilKitError{"Union discriminator is negative"};
+        if (discriminator == 0)
```
VJanKraemer, line 165: Wouldn't this be allowed according to the "spec" in COE-66476? I.e. write the value 0 for the tag to the serialize buffer and then nothing for the actual union value?

VDanielEdwards: Good point. I'll experiment some more on the internal side to see if a union value can be 'cleared' somehow.

VDanielEdwards: And after some further investigation, it turns out that it is actually possible to clear the union explicitly in some (but not all) language bindings on the internal side.

Thank you for the feedback, the check and throw has been removed!

VJanKraemer (approved):
Looks good for me from a barebones serialize/deserialize point of view :) 👍

PR #124, opened 2024-09-27
ci: clean up build workflows

MariusBgm:
Move the runner setup into the workflows and keep the build-cmake-preset
action as simple as possible.
Add workflow_dispatch to mingw so we can run it manually.

### DIFF
--- a/.github/workflows/build-mingw64.yml
+++ b/.github/workflows/build-mingw64.yml
@@ -1,60 +1,45 @@
-name: 'mingw64 release Builds'
+name: 'mingw64-x86_64'
 on:
   workflow_call:
     inputs:
-      do_package:
-        required: false
-        type: boolean
-        default: false
-      do_package_symbols:
-        required: false
-        type: boolean
-        default: false
-      retention_days:
-        required: false
-        type: number
-        default: 14
       run_build:
         required: true
         type: boolean
-
+  workflow_dispatch:
+    inputs:
+      run_build:
+        default: true
 jobs:
-  gcc13-release:
-    name: GCC 13.2 release Builds for mingw64
+  mingw:
+    name: msys2/mingw64 ${{ matrix.builds.arch }}
+    defaults:
+      run:
+        shell:  C:\shells\msys2bash.cmd {0}
+
+    strategy:
+      matrix:
+        builds:
+          - arch: x86_64
+            bin: /mingw64/bin
+          #- arch: i686
+          #  bin: /mingw32/bin
 
     runs-on: windows-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/build-mingw64.yml
```diff
@@ -1,60 +1,46 @@
...
+      - name: install packages
         run: |
-          echo "C:\msys64" >> $GITHUB_PATH
-          echo "C:\msys64\mingw64\bin\" >> $GITHUB_PATH
-      - name: Install Dependencies
-        run: |
-          pacman -S mingw-w64-x86_64-gcc mingw-w64-x86_64-cmake mingw-w64-x86_64-ninja --noconfirm
-          export PATH="/mingw64/bin/:$PATH"
-        shell:  C:\shells\msys2bash.cmd {0}
+          pacman -S mingw-w64-${{ matrix.builds.arch }}-gcc mingw-w64-${{ matrix.builds.arch }}-cmake mingw-w64-${{ matrix.builds.arch }}-ninja --noconfirm
```
VDanielEdwards, line 33: This does install `cmake` in the MSYS2 environment, but ...

.github/workflows/build-mingw64.yml
```diff
@@ -1,60 +1,46 @@
...
-        shell:  C:\shells\msys2bash.cmd {0}
-      - name: Test
-        id: cmake-test-mingw64
-        run: |
-          export PATH="/mingw64/bin/:$PATH"
-          echo "PATH: $PATH"
-          ctest --preset gcc-release --output-on-failure
-        shell:  C:\shells\msys2bash.cmd {0}
+
+      - uses: ./.github/actions/build-cmake-preset
```
VDanielEdwards, line 40: .. this action does not use the MSYS2 environment for the build. The default shell is not passed to the action, it's only used for the workflow itself.

MariusBgm: ah, thanks. that's the magic sauce that I missed :-)

VDanielEdwards (approved):
If the `cmake` executable can be found again in the correct MinGW installation, this PR is ready for merging 👍

VDanielEdwards (approved):

MariusBgm:
> If the `cmake` executable can be found again in the correct MinGW installation, this PR is ready for merging 👍

Yes, the cmake executable is now properly found. 
It helps that we can rely on the fact that there is a bash available on all systems. 
We now export the extra-paths in the bash shell of the composite action. This allows setting the path for the mingw / msys2 installations from outside of the `build-cmake-preset`-action.

PR #128, opened 2024-10-04
dev: Add missing controllerType in ServiceDescriptor of SimulatedController

KonradBkd:
## Subject

Using Trace Logging in the NetSim API caused the exception `supplementalData.size() > 0`

## Description

This was the case because the controllerType (by name) was not saved on the supplemental data of the mirrored controller in the NetSim API when a new controller got discovered. 

With log level Trace, the mirrored controller then caused the exeption when trying to log the controller type in its service descriptor.

Finally, the exception got catched in the hourglass `CAPI_CATCH_EXCEPTIONS`.

Fixed by writing the controller type in the suppl. info map of the mirrored controller. Also made the exception more specific.

- Code Review
- Build this branch, make sure the NetSim uses the SilKit.dll with the fix.
- Run a Demo with NetSim and  `-l trace` on the Netsim.exe (or via trace lvl in the participant config). No exception must appear. 
  
- [x] Changelog.md updated

### DIFF
--- a/SilKit/source/experimental/netsim/SimulatedNetworkInternal.cpp
+++ b/SilKit/source/experimental/netsim/SimulatedNetworkInternal.cpp
@@ -8,6 +8,7 @@
 #include "eventproducers/FlexRayEventProducer.hpp"
 #include "eventproducers/EthernetEventProducer.hpp"
 #include "eventproducers/LinEventProducer.hpp"
+#include "ServiceConfigKeys.hpp"
 
 namespace SilKit {
 namespace Experimental {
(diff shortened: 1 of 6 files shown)


SilKit/source/experimental/netsim/SimulatedNetworkInternal.cpp
```diff
@@ -57,21 +58,33 @@ void SimulatedNetworkInternal::CreateAndSetEventProducer()
     }
 }
 
+auto SimulatedNetworkInternal::ExtractControllerTypeName(const SilKit::Core::ServiceDescriptor& serviceDescriptor)
```
MariusBgm, line 61: since this is only used in `SimulatedNetworkRouter::AddSimulatedController`, i wonder if we could just move it down inside the call of AddSimulatedController.

SilKit/source/experimental/netsim/SimulatedNetworkInternal.cpp
```diff
@@ -57,21 +58,33 @@ void SimulatedNetworkInternal::CreateAndSetEventProducer()
...
-    auto serviceDescriptorStr = serviceDescriptor.to_string();
 
     _controllerDescriptors[fromParticipantName][serviceId] = controllerDescriptor;
 
     auto userSimulatedController = _userSimulatedNetwork->ProvideSimulatedController(controllerDescriptor);
 
     if (userSimulatedController)
     {
-        _simulatedNetworkRouter->AddSimulatedController(fromParticipantName, controllerName, serviceId,
+        std::string controllerTypeName = ExtractControllerTypeName(serviceDescriptor);
```
MariusBgm, line 85: couldn't we make this call inside of AddSimulatedController? we're passing serviceDescriptor there anyways. Or am I overlooking some other usages of this?

MariusBgm (approved):
LGTM! Demo with Network simulator works now.
Just some minor nitpicks, on where to place the new method / calls to it.

SilKit/source/experimental/netsim/SimulatedNetworkInternal.cpp
KonradBkd: I wanted to isolate the "out parameter" and possible execption that happens there just for readability.

PR #132, opened 2024-10-11
SILKIT-1612 log: add participant name to file names

MariusBgm:
Ensure that filenames are (more) unique, just in case multiple log files are opened in quick succession.
In the future we might want to provide some templating for the user to choose the file name.


VDanielEdwards (approved):
𝔏𝔊𝔗𝔐 👍

MariusBgm:
I'll update the documentation string for `Logname`

PR #134, opened 2024-10-14
demos: add an async simstep handler demo

MariusBgm:
a simple demo that shows how to have a worker thread running in parallel to the SIL Kit `SimulationStepHandlerAsync` and `CompleteSimulationStep` calls on the main thread.


KonradBkd:
Such a demo is now included in the Demos/api/Orchestration in https://github.com/vectorgrp/sil-kit/pull/143

PR #137, opened 2024-10-24
cmake: improve platform/compiler/architecture detection and make cross-building to QNX on Windows possible

VDanielEdwards:
## Subject

This PR adapts the platform/compiler/architecture extraction logic to allow cross-builds to QNX on Windows. It also unifies some logic and adds VS2022 detection, with a new fallback for 'unknown' MSVC toolset versions.

It also adapts the QNX toolchain base file to skip settings the rpath on the build-artifacts which is extremely broken on Windows. Since the QNX binaries (e.g., test executables) cannot be executed on the host (_not_ the cross-build target) anyway, this shouldn't break anything.

- [x] Changelog.md updated


SilKit/cmake/get_compiler_arch.cmake
```diff
@@ -80,73 +80,93 @@ function(get_linux_distro outDistroName outDistroVersion)
...
         elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-            set(_id "clang")
-        endif()
-        message(STATUS "Using compiler version: ${_id}-${CMAKE_CXX_COMPILER_VERSION}")
-        set(_tool_tag ${_id})
-    elseif(MINGW)
-        set(_tool_tag "MinGW")
-    elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
-        silkit_validate_preset_toolset()
-        if(CMAKE_VS_PLATFORM_TOOLSET MATCHES "v140")
```
MariusBgm, line 132: I think CMAKE_VS_PLATFORM_TOOLSET was used for very old cmake versions, but I think we're safe to deprecate that.
The `MSVC_TOOLSET_VERSION` is safe to use since cmake Version 3.12

MariusBgm (approved):
LGTM!
Tested on NetBSD, VS2017 Generator and QNX crossbuild on Windows.

SilKit/cmake/get_compiler_arch.cmake
VDanielEdwards: The top-level `CMakeLists.txt` has had the minimum version set to 3.12 for quite a while already.

SilKit/cmake/get_compiler_arch.cmake
```diff
@@ -80,73 +80,93 @@ function(get_linux_distro outDistroName outDistroVersion)
...
-        set(_system_arch "x86")
+        # building on (or for) Windows-on-ARM will lead to invalid architecture values
+
         if("${SYSTEM_BITNESS}" STREQUAL "64")
             set(_system_arch "x86_64")
+        else()
+            set(_system_arch "x86")
         endif()
+    elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "QNX")
+        set(_platform "QNX")
```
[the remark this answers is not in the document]
VDanielEdwards: Ah yes, I added the `elseif` branch before deciding to use the system name as the default. Good catch thanks!

PR #146, opened 2024-11-14
SILKIT-1650: Forward specific SIL Kit exceptions

KonradBkd:
SIL Kit specific exceptions (e.g. `SilKit::ConfigurationError`) are not transported to the C++ side through the hourglass. There, only the exception `Silkit::SilkitError` is thrown (_ThrowOnError.hpp_)

In this PR:
- `SilKit_ReturnCodeXYZ` for all specific exceptions
- The `CAPI_CATCH_EXCEPTIONS` macro catches the specific exception and returns the corresponding `SilKit_ReturnCode_XYZ`
- In _ThrowOnError.hpp_, the return code leads to the specific exception again
- Tests for `CAPI_CATCH_EXCEPTIONS` and `ThrowOnError()` in _Test_CApiExceptions.cpp_
- Fix _Test_CapiSilKit_ which was using outdated arguments in `CreateParticipant`. Also, the silkit_function_mapping test was misleading, as it was actually only testing for bad / invalid parameters with wrong expectations.

### DIFF
--- a/SilKit/include/silkit/capi/Types.h
+++ b/SilKit/include/silkit/capi/Types.h
@@ -43,15 +43,32 @@ typedef struct SilKit_Experimental_SystemController SilKit_Experimental_SystemCo
 
 typedef int32_t SilKit_ReturnCode;
 
-#define SilKit_ReturnCode_SUCCESS ((SilKit_ReturnCode)0)
-#define SilKit_ReturnCode_UNSPECIFIEDERROR ((SilKit_ReturnCode)1)
-#define SilKit_ReturnCode_NOTSUPPORTED ((SilKit_ReturnCode)2)
-#define SilKit_ReturnCode_NOTIMPLEMENTED ((SilKit_ReturnCode)3)
-#define SilKit_ReturnCode_BADPARAMETER ((SilKit_ReturnCode)4)
-#define SilKit_ReturnCode_BUFFERTOOSMALL ((SilKit_ReturnCode)5)
-#define SilKit_ReturnCode_TIMEOUT ((SilKit_ReturnCode)6)
-#define SilKit_ReturnCode_UNSUPPORTEDSERVICE ((SilKit_ReturnCode)7)
-#define SilKit_ReturnCode_WRONGSTATE ((SilKit_ReturnCode)8) // Returned on exception SilKit::StateError (CapiImpl.h)
+#define SilKit_ReturnCode_SUCCESS ((SilKit_ReturnCode)0)               
+#define SilKit_ReturnCode_UNSPECIFIEDERROR ((SilKit_ReturnCode)1)      
+#define SilKit_ReturnCode_NOTSUPPORTED ((SilKit_ReturnCode)2)          
+#define SilKit_ReturnCode_NOTIMPLEMENTED ((SilKit_ReturnCode)3)        
+#define SilKit_ReturnCode_BADPARAMETER ((SilKit_ReturnCode)4)          
+#define SilKit_ReturnCode_BUFFERTOOSMALL ((SilKit_ReturnCode)5)        
+#define SilKit_ReturnCode_TIMEOUT ((SilKit_ReturnCode)6)               
+#define SilKit_ReturnCode_UNSUPPORTEDSERVICE ((SilKit_ReturnCode)7)    
+
+// The following return codes have an corresponding specific SIL Kit exception.
+// If an error occurs and a specific exception is thrown, it is catched in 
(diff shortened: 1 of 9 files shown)


SilKit/source/capi/Test_CapiExceptions.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+namespace {
+
+class Test_CapiExceptions : public testing::Test
+{
+public:
+    Test_CapiExceptions() {}
+};
+
+SilKit_ReturnCode throw_SilKitError()
```
VDanielEdwards, line 35: You could merge all of these into a single function template, like so:

```c++
template<typename T>
auto TestExceptionToErrorCode()
try
{
    throw T{"error msg"};
}
CAPI_CATCH_EXCEPTIONS
```

And invoke it like so:

```c++
EXPECT_EQ(TestExceptionToErrorCode<SilKit::LengthError>(), SilKit_ReturnCode_LENGTH_ERROR);
```

SilKit/include/silkit/capi/Types.h
```diff
@@ -43,15 +43,32 @@ typedef struct SilKit_Experimental_SystemController SilKit_Experimental_SystemCo
...
+#define SilKit_ReturnCode_UNSPECIFIEDERROR ((SilKit_ReturnCode)1)      
+#define SilKit_ReturnCode_NOTSUPPORTED ((SilKit_ReturnCode)2)          
+#define SilKit_ReturnCode_NOTIMPLEMENTED ((SilKit_ReturnCode)3)        
+#define SilKit_ReturnCode_BADPARAMETER ((SilKit_ReturnCode)4)          
+#define SilKit_ReturnCode_BUFFERTOOSMALL ((SilKit_ReturnCode)5)        
+#define SilKit_ReturnCode_TIMEOUT ((SilKit_ReturnCode)6)               
+#define SilKit_ReturnCode_UNSUPPORTEDSERVICE ((SilKit_ReturnCode)7)    
+
+// The following return codes have an corresponding specific SIL Kit exception.
+// If an error occurs and a specific exception is thrown, it is catched in
```
MariusBgm, line 56:
```suggestion
// If an error occurs and a specific exception is thrown, it is caught in 
```

SilKit/include/silkit/capi/Types.h
```diff
@@ -43,15 +43,32 @@ typedef struct SilKit_Experimental_SystemController SilKit_Experimental_SystemCo
...
+
+#define SilKit_ReturnCode_WRONGSTATE ((SilKit_ReturnCode)8)             // SilKit::StateError
+#define SilKit_ReturnCode_TYPECONVERSION_ERROR ((SilKit_ReturnCode)9)   // SilKit::TypeConversionError
+#define SilKit_ReturnCode_CONFIGURATION_ERROR ((SilKit_ReturnCode)10)   // SilKit::ConfigurationError
+#define SilKit_ReturnCode_PROTOCOL_ERROR ((SilKit_ReturnCode)11)        // SilKit::ProtocolError
+#define SilKit_ReturnCode_ASSERTION_ERROR ((SilKit_ReturnCode)12)       // SilKit::AssertionError
+#define SilKit_ReturnCode_EXTENSION_ERROR ((SilKit_ReturnCode)13)       // SilKit::ExtensionError
+#define SilKit_ReturnCode_LOGIC_ERROR ((SilKit_ReturnCode)14)           // SilKit::LogicError
+#define SilKit_ReturnCode_LENGTH_ERROR ((SilKit_ReturnCode)15)          // SilKit::LengthError
+#define SilKit_ReturnCode_OUTOFRANGE_ERROR ((SilKit_ReturnCode)16)      // SilKit::OutOfRangeError
```
VDanielEdwards, line 71: The naming of the new enum values seems a bit inconsistent to me. Our current convention (by example) seems to be to use only uppercase letters and with no underscores between words. But e.g., `OutOfRangeError` is translated to a 'mixed' name `OUTOFRANGE_ERROR` which looks strange to me.

KonradBkd: You're right, UNSPECIFIEDERROR already setted a convention there. Will change...

KonradBkd: Ah now I'm unsure when writing this out. It's funny how the perception immediately jumps on LOG and ICE when reading SilKit_ReturnCode_LOGICERROR. 
Will change it though, consistency should win.

VDanielEdwards (approved):
LGTM!

PR #147, opened 2024-11-14
bump version to 4.0.55

KonradBkd:
Bump SilKitVersion.cmake + Changelog to 4.0.55

### DIFF
--- a/SilKit/cmake/SilKitVersion.cmake
+++ b/SilKit/cmake/SilKitVersion.cmake
@@ -27,7 +27,7 @@
 macro(configure_silkit_version project_name)
     set(SILKIT_VERSION_MAJOR 4)
     set(SILKIT_VERSION_MINOR 0)
-    set(SILKIT_VERSION_PATCH 54)
+    set(SILKIT_VERSION_PATCH 55)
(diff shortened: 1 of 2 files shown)


VDanielEdwards (approved):

VDanielEdwards:
We should add this mechanism as the last step of a successful delivery. We in fact even automate the PR creation. I'll create an internal issue for this.

MariusBgm:
> We should add version bump as the last step of a successful delivery. We could even automate the PR creation. I'll create an internal issue for this.

I second this. should be easy enough to automate

VDanielEdwards:
> > > We should add version bump as the last step of a successful delivery. We could even automate the PR creation. I'll create an internal issue for this.
> > 
> > 
> > I second this. should be easy enough to automate
> 
> Downside is that we have to guess that the next version is a _Patch_ increment. If a some other increment happens, we have to adjust. Looking back, this is not a real problem though.

We could also put the bump into a separate `workflow_dispatch` workflow, which can be parametrized with an option to select the kind of bump. This workflow would be able to see the same secrets as the tagging workflow.

PR #150, opened 2024-11-20
c-api: turn const global into define

VDanielEdwards:
## Subject

When linking two separately compiled object files, both using the C-API, the linker informed me about a multiply defined symbol. This turned out to be the `const` global `SilKit_LinDataLengthUnknown`.

I did grep through the C headers and haven't found another one of those.

Neither the `.dll` nor the `.so` export this symbol so no trouble there.

Issue: SILKIT-1654

- [ ] Changelog.md updated

### DIFF
--- a/SilKit/include/silkit/capi/Lin.h
+++ b/SilKit/include/silkit/capi/Lin.h
@@ -161,7 +161,7 @@ typedef uint8_t SilKit_LinFrameStatus;
 typedef uint8_t SilKit_LinDataLength;
 
 //! \brief If configured for reception with this value, the data length validation of incoming frames is skipped.
-const SilKit_LinDataLength SilKit_LinDataLengthUnknown = 255u;
+#define SilKit_LinDataLengthUnknown ((SilKit_LinDataLength)255)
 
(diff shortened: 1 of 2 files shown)


SilKit/include/silkit/capi/Lin.h
```diff
@@ -161,7 +161,7 @@ typedef uint8_t SilKit_LinFrameStatus;
 typedef uint8_t SilKit_LinDataLength;
 
 //! \brief If configured for reception with this value, the data length validation of incoming frames is skipped.
-const SilKit_LinDataLength SilKit_LinDataLengthUnknown = 255u;
```
MariusBgm, line 164: oh boy. what an oversight....
is this symbol by default hidden? it's not visible/dllexported, right?
so recompiling with this fix should be fine for all users.

MariusBgm (approved):
looks good, as this seems to only affect a non-visible symbol.
I just tested your changes on ubuntu and Windows with a modified SilKitDemoLin (added a use of the symbol) executable. it doesn't complain about missing symbols.

PR #162, opened 2024-12-17
capi: fix function pointer arguments in declaration

MariusBgm:
As reported by a user of the C-API, the pointer typedefs don't match the actual declarations.


VDanielEdwards (approved):
LGTM! 👍

PR #163, opened 2024-12-20
dev: String definitions for SIL Kit error return codes

KonradBkd:
- Completed C strings in SilKit\source\capi\CapiUtils.cpp
- More clear exception message formatting
- Replaced `SILKIT` with `SIL Kit`


VDanielEdwards (approved):
LGTM! Good finding with the error code strings!

PR #165, opened 2025-01-07
GetParticipantName and GetRegistryUri

KonradBreitsprecherBkd:
# Add dedicated participant string getters

## Summary
This change introduces two dedicated C-API functions:

- `SilKit_Participant_GetParticipantName(void* outParameterValue, size_t* inOutParameterValueSize, SilKit_Participant* participant)`
- `SilKit_Participant_GetRegistryUri(void* outParameterValue, size_t* inOutParameterValueSize, SilKit_Participant* participant)`

and their C++ equivalents in  IParticipant

- `virtual auto GetParticipantName() const -> std::string = 0`
- `virtual auto GetRegistryUri() const -> std::string = 0`

In addition, string copy/size handling was centralized in the C-API layer and tests were updated/extended accordingly.
Further, the new API is used in the sil-kit-system-controller implementation to dynamically retrieve the participant name (set via command line and/or participant configuration). The name has to be added to the `required participant names` of the `workflow configuration`. Without the GetParticipantName functionality, passing a new name by config actually failed because there was no way to retrieve the name from the config programatically.

## Motivation
The generic `SilKit_Participant_GetParameter` API in the first draft depended on an enum that looked like it could be expanded to non-string types but actually did not support this. The new design is more simple and exposes the only two relevant participant string parameters directly and clearly. Future parameter queries could be covered by more explicit getters or by retrieving the current participant configuration.

## Further Notes
 - `CapiParticipant.cpp | CopyStringToOutBuffer`: string size query and nulltermination handling
- `ParameterProvider.hpp | QueryString(SilKit_Participant*, StringQueryFunction)`: Reusable helper to atomically obtain the string that might change between buffer allocation and query.
- Unit and Integration tests are in place
- Docs are still missing until API is stable


SilKit/include/silkit/participant/IParticipant.hpp
```diff
@@ -99,6 +101,9 @@ class IParticipant
 
     //! \brief Return the ILogger at this SIL Kit participant.
     virtual auto GetLogger() -> Services::Logging::ILogger* = 0;
+
+    //! \brief Get a parameter set by an API call and/or the participant configuration.
+    virtual auto GetParameter(SilKit::Parameter parameter) -> const std::string& = 0;
```
VDanielEdwards, line 106: I would produce a new `std::string` on each invocation. This API shouldn't be called more than once (under normal circumstances) anyway. This will also simplify the implementation, because the caching in the Hourglass-API can be dropped.

```suggestion
    virtual auto GetParameter(SilKit::Parameter parameter) -> std::string = 0;
```

SilKit/include/silkit/capi/Participant.h
```diff
@@ -80,6 +81,24 @@ SilKitAPI SilKit_ReturnCode SilKitCALL SilKit_Participant_GetLogger(SilKit_Logge
...
+ * \param outParameterValue The string value of the parameter (out parameter).
+ * \param parameter The parameter to get.
+ * \param participant The participant to get the parameter from.
+ *
+ * Returns the current value of the given parameter. 
+ * Useful for parameters that are passed to the participant via the API and the participant configuration.
+ */
+SilKitAPI SilKit_ReturnCode SilKitCALL SilKit_Participant_GetParameter(const char** outParameterValue,
+                                                                       SilKit_Parameter parameter,
+                                                                       SilKit_Participant* participant);
```
VDanielEdwards, line 96: If we return a single `const char *` (which is owned by the SIL Kit participant and not to be deallocated by the user) then we have to ensure that the string this is pointing to never changes / gets reallocated. This will work for the current set of parameters (participant name and registry uri won't change), but who known if we want to provide parameters that could change in the future.

I would propose an API that copies the strings to user-provided buffer in a similar manner to `snprintf`. It must be allowed to specify the data-pointer as `nullptr` which will only fill in the size (which can then be used to create an appropriately sized buffer).

Same as with the C++ API, I think copying is fine here - these parameters shouldn't be performance critical or queried in hot-loops.

```suggestion
SilKitAPI SilKit_ReturnCode SilKitCALL SilKit_Participant_GetParameter(char* outParameterValueData,
                                                                       size_t *outParameterValueSize,
                                                                       SilKit_Parameter parameter,
                                                                       SilKit_Participant* participant);
```

KonradBkd: Ok I see the point now. With my initial proposal, the cached value behind `const char** outParameterValue` is bound to the participant via the `_parameterProvider` and will leave a dangling pointer if the parameter is deleted.

So the usage would look like:
```
size_t paramSize;
SilKit_Participant_GetParameter(NULL, &paramSize, SilKit_Parameter_ParticipantName, participant);
char* paramValue = malloc(sizeof(char) * paramSize);
// Hope that the parameter size has not changed in the meantime
SilKit_Participant_GetParameter(paramValue, paramSize, SilKit_Parameter_ParticipantName, participant);
// Print paramValue
free(paramValue)
```

What happens if the parameter has changed beween the two calls? Do we actually need a two-step process like:

```
size_t paramSize;
uint64_t parameterRequestID;
// First call already caches the parameterValue (and size) internally and hands over a request ID
SilKit_Participant_RequestParameter(&paramSize, &parameterRequestID, SilKit_Parameter_ParticipantName, participant); 
char* paramValue = malloc(sizeof(char) * paramSize);
// Second call copies the cached value to the user buffer
SilKit_Participant_GetParameter(paramValue, parameterRequestID, participant);
// Print paramValue
free(paramValue)
```

VDanielEdwards: Even if the buffer-pointer isn't `nullptr`, the 'required' size of the buffer will still be returned through the size-pointer. Therefore the user should check that after the second call (which actually copied the data) the size still matches what was expected before.

Some example code in C++ since it is a little bit less messy than C.

```c++
size_t size;
/* check return code */ ... = SilKit_Participant_RequestParameter(nullptr, &size, ...);

std::vector<char> buffer;
while (size != buffer.size())
{
  buffer.resize(size);
  /* check return code */ ... = SilKit_Participant_RequestParameter(buffer.data(), &size, ...);
}
```

Since `buffer.size()` starts out as 0, the loop will do zero iterations if the initial `size` was 0.

If the returned `size` has changed between the two calls, the buffer will be resized accordingly in the next iteration. This repeats for as long as neccessary.

VDanielEdwards: Of course, the function must never actually write more than what is specified through the size-pointer.

VDanielEdwards: Alternatively we could not copy in the case that the user-specified buffer size is too short, and return a special error code. I believe this is how the Win32 string-encoding translation functions are designed.

Personally I like the loop approach because it is very simple, and makes it possible to get a prefix of a certain length without jumping through hoops.

VDanielEdwards: > Even if the buffer-pointer isn't `nullptr`, the 'required' size of the buffer will still be returned through the size-pointer. Therefore the user should check that after the second call (which actually copied the data) the size still matches what was expected before.
> 
> Some example code in C++ since it is a little bit less messy than C.
> 
> ```c++
> ...
> ```
> 
> Since `buffer.size()` starts out as 0, the loop will do zero iterations if the initial `size` was 0.
> 
> If the returned `size` has changed between the two calls, the buffer will be resized accordingly in the next iteration. This repeats for as long as neccessary.

The example code can be optimized by checking for `size > buffer.size()` and putting an additional `buffer.resize(size)` after the loop. This makes 'shrinking' size values more efficient.

KonradBreitsprecherBkd:
I committed an updated version with the suggested pattern.
The `inOutParameterValueSize` includes null-termination, as this is what the user has to allocate.
Not sure about the `inOut` prefix, I think this is the first appearance of a two-way argument in the C API.

```
SilKit_ReturnCode SilKitCALL SilKit_Participant_GetParameter(char* outParameterValue, size_t* inOutParameterValueSize,
                                                             SilKit_Parameter parameter,
                                                             SilKit_Participant* participant)
try
{
    ASSERT_VALID_OUT_PARAMETER(inOutParameterValueSize);
    ASSERT_VALID_POINTER_PARAMETER(participant);

    auto cppParticipant = reinterpret_cast<SilKit::IParticipant*>(participant);
    auto cppParameter = static_cast<SilKit::Parameter>(parameter);
    auto parameterValue = cppParticipant->GetParameter(cppParameter);

    // outParameterValue == nullptr indicates a size-check only, otherwise copy
    if (outParameterValue != nullptr)
    {
        size_t sizeToCopy;
        if (*inOutParameterValueSize >= parameterValue.size() + 1)
        {
            // Don't copy more than we actually have
            sizeToCopy = parameterValue.size();
        }
        else
        {
            // Don't copy more than the given size
            sizeToCopy = *inOutParameterValueSize - 1;
        }
        parameterValue.copy(outParameterValue, sizeToCopy);
        outParameterValue[sizeToCopy] = '\0';
    }
    *inOutParameterValueSize = parameterValue.size() + 1;
    return SilKit_ReturnCode_SUCCESS;
}
CAPI_CATCH_EXCEPTIONS
```

Usage in C++: 

```
auto ParameterProvider::GetParameter(SilKit_Participant* participant, Parameter parameter) -> std::string
{
    std::vector<char> buffer;
    size_t size = 0;
    SilKit_Parameter cParameter = static_cast<SilKit_Parameter>(parameter);
    {
        const auto returnCode = SilKit_Participant_GetParameter(nullptr, &size, cParameter, participant);
        ThrowOnError(returnCode);
    }
    while (size > buffer.size())
    {
        buffer.resize(size);
        const auto returnCode = SilKit_Participant_GetParameter(buffer.data(), &size, cParameter, participant);
        ThrowOnError(returnCode);
    }
    buffer.resize(size);

    return std::string{buffer.data()};
}
```

SilKit/include/silkit/detail/impl/participant/ParameterProvider.hpp
VDanielEdwards, line 1: For future reference, this class can be removed and the implementation moved to the participant implementation.

SilKit/include/silkit/capi/Participant.h
```diff
@@ -80,6 +81,28 @@ SilKitAPI SilKit_ReturnCode SilKitCALL SilKit_Participant_GetLogger(SilKit_Logge
...
+ * \param outParameterValue A buffer to copy the null-terminated parameter value to. 
+                            Passing a nullptr is valid and indicates a size-check via inOutParameterValueSize to allocate the buffer.
+ * \param inOutParameterValueSize The size of the parameter including null-termination. 
+ * \param parameter The parameter to get.
+ * \param participant The participant to get the parameter from.
+ *
+ * Returns the current value of the given parameter. 
+ * Useful for parameters that are passed to the participant via the API and the participant configuration.
+ */
+SilKitAPI SilKit_ReturnCode SilKitCALL SilKit_Participant_GetParameter(char* outParameterValue,
```
VDanielEdwards, line 96: The `outParameterValue` should be a `void *`. This enables non-string parameters, e.g., byte blobs.

## Trailing nul character

I'm also against both, writing the trailing nul character, and including it in the output length.
If the trailing nul character is required for other APIs, the caller must simply allocate one more byte and can explicitly set that to nul after the `GetParameter` call.

SilKit/include/silkit/participant/parameters.hpp
```diff
@@ -0,0 +1,23 @@
...
+// note: never reuse old numbers!
+//! \brief Available parameters to query
+enum class Parameter : SilKit_Parameter
+{
+    //! An undefined parameter
+    Undefined = SilKit_Parameter_Undefined,
+    //! The name of the participant
+    ParticipantName = SilKit_Parameter_ParticipantName,
+    //! The registry URI
+    RegistryUri = SilKit_Parameter_ReistryUri,
```
KonradBreitsprecherBkd: Typo

SilKit/include/silkit/capi/Participant.h
```diff
@@ -63,6 +63,37 @@ SilKitAPI SilKit_ReturnCode SilKitCALL SilKit_Participant_GetLogger(SilKit_Logge
...
+                                                                                SilKit_Participant* participant);
+
+/*! \brief Retrieve the registry URI from a participant.
+ *
+ * \param outParameterValue A buffer to copy the null-terminated registry URI to.
+ *                          Passing a nullptr is valid and indicates a size-check via inOutParameterValueSize.
+ * \param inOutParameterValueSize The required/provided size of the value including null-termination.
+ * \param participant The participant to query.
+ */
+SilKitAPI SilKit_ReturnCode SilKitCALL SilKit_Participant_GetGetRegistryUri(void* outParameterValue,
```
VDanielEdwards:
```suggestion
SilKitAPI SilKit_ReturnCode SilKitCALL SilKit_Participant_GetRegistryUri(void* outParameterValue,
```

Lol 🐱

SilKit/include/silkit/detail/impl/participant/ParameterProvider.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+namespace SilKit {
+DETAIL_SILKIT_DETAIL_VN_NAMESPACE_BEGIN
+namespace Impl {
+
+ParameterProvider::ParameterProvider()
+{
+}
+
+auto ParameterProvider::QueryString(SilKit_Participant* participant,
```
VDanielEdwards: I wouldn't make this a dedicated class. I would just include the `QueryString` function as a private method in the `Participant` class in the hourglass implementation, or even as a free function in the impl namespace.

SilKit/include/silkit/capi/Participant.h
KonradBreitsprecherBkd: :) Das schieb ich auf die Hitze

SilKit/include/silkit/capi/Participant.h
```diff
@@ -63,6 +63,37 @@ SilKitAPI SilKit_ReturnCode SilKitCALL SilKit_Participant_GetLogger(SilKit_Logge
...
 
+
+/*! \brief Retrieve the participant name from a participant.
+ *
+ * \param outParameterValue A buffer to copy the null-terminated participant name to.
+ *                          Passing a nullptr is valid and indicates a size-check via inOutParameterValueSize.
+ * \param inOutParameterValueSize The required/provided size of the value including null-termination.
+ * \param participant The participant to query.
+ */
+SilKitAPI SilKit_ReturnCode SilKitCALL SilKit_Participant_GetParticipantName(void* outParameterValue,
```
VDanielEdwards: Since the functions aren't 'generic' anymore (i.e., always write strings of `char` into the output buffer), you can now use the concrete type: `char *` instead of `void *`.

SilKit/include/silkit/detail/impl/participant/ParameterProvider.hpp
KonradBreitsprecherBkd: You mean to get rid of the whole `ParameterProvider` class?

SilKit/source/capi/CapiParticipant.cpp
```diff
@@ -15,11 +15,35 @@
...
 
+#include <algorithm>
 #include <memory>
 #include <map>
 #include <mutex>
 #include <fstream>
 
+namespace
+{
+auto CopyStringToOutBuffer(const std::string& value, void* outParameterValue, size_t* inOutParameterValueSize)
```
VDanielEdwards: Since we're now compiling with glorious C++17 you can switch the type of `value` to `std::string_view` (and include the `<string_view>` header). This will ensure that there is no copy made for the `value` parameter (and doesn't bind us to `std::string` internally).

SilKit/source/core/mock/participant/MockParticipant.hpp
```diff
@@ -349,6 +349,15 @@ class DummyParticipant : public IParticipantInternal
         return &logger;
     }
 
+    auto GetParticipantName() const -> std::string override
```
VDanielEdwards, line 352: Its probably not a huge issue (because the method is hopefully only called once), but this will lead to at least two 'throw-away' copies (due to the return here) of the name in the implementation of the `C` API function. The `C` API implementation should probably use an 'internal-only' API of the participant that does not copy the string on each call, but returns a reference (or `std::string_view`).

SilKit/include/silkit/detail/impl/participant/ParameterProvider.hpp
VDanielEdwards: Yeah pretty much. It doesn't really do anything but provide wrappers, or am I missing something?

KonradBreitsprecherBkd: No storage, no reason to have more than a single instance.. will remove the class

VDanielEdwards (approved):

MariusBgm:
i think we should consider this in a bigger context of providing somthing like service discovery. Not sure if we could add this to a SystemMonitor or some other new Service Entry point, e.g. something stably wrapping ServiceDiscovery.
At the very least, i can think of a `SilKit_Participant_GetSimulationName`

KonradBreitsprecherBkd:
> i think we should consider this in a bigger context of providing somthing like service discovery. Not sure if we could add this to a SystemMonitor or some other new Service Entry point, e.g. something stably wrapping ServiceDiscovery. At the very least, i can think of a `SilKit_Participant_GetSimulationName`

The underlying problem is more general: The user currently cannot retreive it's actual (final) participant configuration merged from API calls and a passed Yaml config. Aside from the name and uri, this problem exists for all arguments of `CreateController(...)` calls that can be overwritten by config. This part is completely local, so I don't see this as part of the service discovery. I agree that this is not a general solution, but simply for uri and name. A general solution would be sth like `auto config=participant.getConfiguration()` and then work on the  `config` object to extract the values. However for the central property of the participant name, I would expect the API to have a dedicated getter. 

The `SimulationName` is another case, currently it's encoded in the URI, right?

MariusBgm:
> > i think we should consider this in a bigger context of providing somthing like service discovery. Not sure if we could add this to a SystemMonitor or some other new Service Entry point, e.g. something stably wrapping ServiceDiscovery. At the very least, i can think of a `SilKit_Participant_GetSimulationName`
> 
> The underlying problem is more general: The user currently cannot retreive it's actual (final) participant configuration merged from API calls and a passed Yaml config. Aside from the name and uri, this problem exists for all arguments of `CreateController(...)` calls that can be overwritten by config. This part is completely local, so I don't see this as part of the service discovery. I agree that this is not a general solution, but simply for uri and name. A general solution would be sth like `auto config=participant.getConfiguration()` and then work on the `config` object to extract the values. However for the central property of the participant name, I would expect the API to have a dedicated getter.
> 
First, I like easy APIs for easy stuff, like this. 
However, that was intended by the original design: participants cannot get the current participant name or network name of controllers (think "portable, pre-built participant executables").
Would we add GetNetworkName to controllers, too?

> The `SimulationName` is another case, currently it's encoded in the URI, right?

yes. but my point was, should we create a dedicated API to **get the current values from a participant, like participant name, registry uri, simulation name**?  

Moving away from the old design of not allowing introspection:
a.) either,  ad hoc additions of methods to the toplevel participant object, or 
b.) create a dedicated, coherent API for introspection (like an enhanced public serviceDiscovery or systemMonitor with participant name, registry uri, simulation name etc).

snps-paul:
> A general solution would be sth like `auto config=participant.getConfiguration()` and then work on the `config` object to extract the values.

I agree with this. As I understand it, irrespective of whether the user provides an explicit config file or not, the `Participant` always stores the 'final' parameters in an internal `config` object. In that case, I think it makes sense to expose that `config` object (with the relevant methods for configurable params: `GetParticipantName()`, `GetRegistryUri()`, ... `GetConfigByCanControllerName(controllerName).GetNetworkName()`, ..., etc.)  in order to query the 'final' values. Then as already mentioned, the user could get the parameters from the `config` obtained via `auto config=participant.getConfiguration()`. 

I would suggest to go with the generic solution to solve the core issue, or is there a reason why that is not desirable at the moment?

>b.) create a dedicated, coherent API for introspection (like an enhanced public serviceDiscovery or systemMonitor with participant name, registry uri, simulation name etc).

Not sure if we need a dedicated introspection API for this but if we go in this direction then I think it would be simpler if the introspection API allows fetching these 'static/pre-simulation' config/API parameters without requiring the user to start any service/simulation.

SilKit/source/capi/Test_CapiGetParameter.cpp
```diff
@@ -0,0 +1,106 @@
...
+TEST_F(Test_CapiGetParameter, getparticipantname_truncates_and_still_null_terminates)
+{
+    auto cMockParticipant = (SilKit_Participant*)&mockParticipant;
+
+    char buffer[5]{};
+    size_t bufferSize = sizeof(buffer);
+    const auto returnCode = SilKit_Participant_GetParticipantName(buffer, &bufferSize, cMockParticipant);
+
+    EXPECT_EQ(returnCode, SilKit_ReturnCode_SUCCESS);
+    EXPECT_EQ(buffer[sizeof(buffer) - 1], '\0');
```
snps-paul, line 103: I think we should compare  the truncated string `"Mock"` with the contents of `buffer` to verify truncation. The `\0` position that we are checking is not conclusive that copy and truncation occurred as it is in the same position since `buffer` was initialized.

PR #166, opened 2025-01-14
demos: benchmark demo with pubsub fully connected

VJonasHolley:
So far, a ring topology is implemented for pubsub communication in the benchmark demo. Since this introduces an inconsistency with ethernet and can communication, a fully connected network is chosen instead for the pubsub case.

Moreover, the total number of messages within an entire simulation is determined a priori based on the simulation parameters. Previously, a counter has been increased on every message reception.

- [ ] Changelog.md updated


KonradBkd (approved):
Looks good!

PR #170, opened 2025-01-21
demos: Add --sleep command line option

KonradBkd:
For some scenarios doing a `std::this_thread::sleep_for` in the WorkerThread  (async) / SimStepHandler (sync) is useful in the demos:

- When testing the Soft/HardResponseTimeout 
- To slow down the execution without adding the wall clock coupling to a participant configuration

Here, a command line option to configure a sleep in the WorkerThread / SimStepHandler is introduced.
```
-s, --sleep <ms>             | The sleep duration per work cycle in milliseconds.
                               Default is no sleeping.
                               Using this options overrides the default execution slow down.
                               Cannot be used together with '--fast'.

```

Additionally, the participant config samples for remote logging now use log level info.


VJanKraemer (approved):
Tested and looks good :)

PR #171, opened 2025-01-21
SILKIT-1676: Block participants from joining the same simulation multiple times

VDanielEdwards:
## Subject

The first attempt to connect to a simulation where a participant with the same name was present was blocked, but subsequent attempts were not blocked.

This change fixes this behavior, by checking that the disconnecting peer is actually the peer that connected initially before removing the connection information on peer shutdown.

- [x] Changelog.md updated


KonradBkd (approved):
Checking `connectedParticipant->peer != peer` makes sense, LGTM

PR #172, opened 2025-01-27
SILKIT-1667: Fix hop on with wall clock coupling

KonradBkd:
Before, participants with wall clock coupling (enabled via `Experimental | TimeSynchronization | AnimationFactor`) would wait to catch up on the elapsed time if joining a already running simulation with time sync. (aka hop-on). 

Now, the thread that handles the wall clock coupling is only started after the evaluation if it is a hop-on scenario or not. If so, the already elapsed simulation time is used as an offset for the wall clock coupling.


VJanKraemer (approved):
Tested locally and looks good to me 👍

PR #175, opened 2025-01-29
dev: Allow to explicitly reset the static signal handler

KonradBkd:
The worker thread of the static signal handler (holding user callbacks) is potentially destruced after the participant. TSAN runs showed use-after-free problems in the `sil-kit-system-controller` when using CTRL-C on it in a running simulation (EthernetDemo). 

Here a explicit function `SilKit::Util::ShutdownSignalHandler` is introduced that destroys the `gSignalMonitor`. Using this at the end of the `sil-kit-system-controller` resolves the TSAN issue. Same is applied to the `ApplicationBase` of the communication demos that also uses the signal handler.

### DIFF
--- a/Demos/communication/include/SignalHandler.hpp
+++ b/Demos/communication/include/SignalHandler.hpp
@@ -255,6 +255,7 @@ void RegisterSignalHandler(SignalHandler handler)
 {
     gSignalMonitor.reset(new SignalMonitor(std::move(handler)));
 }
+
 void ShutdownSignalHandler()
 {
     gSignalMonitor.reset();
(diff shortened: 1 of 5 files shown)


Demos/communication/include/SignalHandler.hpp
```diff
@@ -255,3 +255,7 @@ void RegisterSignalHandler(SignalHandler handler)
 {
     gSignalMonitor.reset(new SignalMonitor(std::move(handler)));
 }
+void ShutdownSignalHandler()
```
VDanielEdwards, line 258: Nur noch eine Leerzeile hier 👾 

```suggestion

void ShutdownSignalHandler()
```

PR #178, opened 2025-02-04
SILKIT-1691: Prevent creation of multiple RPCServerInternal for same ClientUUID

KonradBkd:
In a specific label setup, an additional `RPCServerInternal` is incorrectly created. This means that with one client call you suddenly get two calls to the RPC server's handler.

In the fix here, a map for `internalRpcServers` is used for  the RPCServer and thus multiple creation of `RPCServerInternals` for the same ClientUUID is prevented.

### DIFF
--- a/SilKit/IntegrationTests/ITest_LabelsMatching.cpp
+++ b/SilKit/IntegrationTests/ITest_LabelsMatching.cpp
@@ -1,23 +1,6 @@
-/* Copyright (c) 2022 Vector Informatik GmbH
-
-Permission is hereby granted, free of charge, to any person obtaining
-a copy of this software and associated documentation files (the
-"Software"), to deal in the Software without restriction, including
-without limitation the rights to use, copy, modify, merge, publish,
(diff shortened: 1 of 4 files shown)


VDanielEdwards (changes requested):

VDanielEdwards (approved):
💯 LGTM!

SilKit/IntegrationTests/ITest_LabelsMatching.cpp
```diff
@@ -0,0 +1,211 @@
...
+// SPDX-License-Identifier: MIT
+
+#include <memory>
+#include <thread>
+#include <string>
+#include <chrono>
+#include <iostream>
+#include <atomic>
+
+#include <unordered_map> //remove this after rebase on cmake-cleanup-branch
```
VDanielEdwards, line 12: ? 😆

KonradBkd: Removed :)

PR #179, opened 2025-02-04
fix docs for benchmark demo

VJonasHolley:
## Description
- Aligned optional and positional command line arguments.
- Docs update (simulation step size, service type as command line arguments).


KonradBkd (approved):
LGTM

PR #183, opened 2025-02-11
SILKIT-1623: Performance tests

VJonasHolley:
## Subject
- Delete old performance tests 
- Add new python script for testing performance

- Is it reasonable to explicitly set a compiler in the CMake generation? The compiler choice is strongly dependent on the individual setup.
- To be tested: Does the script work under Windows? Does it need to?


VDanielEdwards:
@VJonasHolley, would you mind if I edited the new script such that it is a bit more PEP 8 compliant? Also which pypi package does the `from git import Repo` refer to?

VJonasHolley:
@VDanielEdwards, it would be great if you could adapt the script!
The `from git import Repo` refers to the GitPython package, see https://gitpython.readthedocs.io/en/stable/index.html.

KonradBkd:
Really nice, this will be quite useful for all kinds of performance related work!

@VDanielEdwards @VJonasHolley 
I made some modifications to the script:
- Runs on Windows now
- Removed threading, `subprocess.Popen` already can run in the background
- Use argparse for command line parsing
- Add requrements.txt for the required packages

The parameters for benchmark/latency are currently reduced that the script runs fast during developlent (see TODOs in the code).

@VDanielEdwards Feel free to go over the script again
@VJonasHolley After the refactoring, could you have a look at the PASSED/FAILED logic? If something goes wrong with the reference (e.g. it gives bad results), the script will PASS. Like with the current 'debug' paramter set I get:
```
----- Test Report (start) -----

Latency:                     PASSED with 24.3075 us    (reference value: 45.67100000000001 us).
Throughput (large messages): PASSED with 1015.22 MiB/s (reference value: 348.09399999999994 MiB/s).
Throughput (small messages): PASSED with 1.96016 MiB/s (reference value: 1.685072 MiB/s). 
```
This "too good to be true result" should also be a fail.

VDanielEdwards:
@KonradBkd @VJonasHolley If might have overdone it a bit, but the test set is now configurable using a config file 😅

KonradBkd (approved):
@VDanielEdwards Really nice update👍
I restored the original parameters for the production runs, ready to merge from my side.

VJonasHolley:
@KonradBkd, @VDanielEdwards: Great script improvements! 🚀

I added some minor modifications:
- Read the latest kpi_value from the under-test data (otherwise the test result won't change when the test is repeated)
- Issue a warning if the standard deviation of the under-test data is "too large" (which is based on a heuristic right now)
- Increase the ``--message-count`` parameter in the throughput case in order to match the results of the SIA paper

Issue #188, opened 2025-03-06
Calling Convention Changes Affecting Multiple Files


TPfitzer:
### Description:
Calling convention affect the following files and functions:

**DataSubscriber.hpp** - Method: TheDataMessageHandler
**RpcClient.hpp** - Method: TheRpcCallResultHandler
**RpcServer.hpp** - Method: TheRpcCallHandler
**ThrowOnError.hpp** - Method: ThrowOnError

These methods impact compatibility and require adjustments in the affected components. Further review and potential fixes are needed.

**Steps to Reproduce:**
Change Calling Convention to __stdcall. I assume the compiler will fetch the default calling convention for this methods if no dedicated convention is specified

**Local Solution:**
I fixed the issue locally by adding the Makro SilKitCALL

VDanielEdwards:
Thank you for the report (and the fix)!

VDanielEdwards:
Just for clarification, what is the issue with the `ThrowOnError` function? It is a free function and is not (?) used as a callback through the C-API.

TPfitzer:
@VDanielEdwards you are correct, sorry for the overshoot. Maybe i just changed it in the flow. I just rechecked and its fine without adding calling convention.

VDanielEdwards:
@TPfitzer Thank you for the clarification. We'll fix this ASAP.

PR #189, opened 2025-03-06
hourglass: calling convention of trampoline functions (#188)

VDanielEdwards:
## Subject

Adds the calling convention macro on the trampoline functions of the hourglass API. This was reported in issue #188.

- [ ] Changelog.md updated

### DIFF
--- a/SilKit/include/silkit/detail/impl/services/pubsub/DataSubscriber.hpp
+++ b/SilKit/include/silkit/detail/impl/services/pubsub/DataSubscriber.hpp
@@ -22,6 +22,7 @@
 #pragma once
 
 #include "silkit/capi/DataPubSub.h"
+#include "silkit/capi/SilKitMacros.h"
 
 #include "silkit/services/pubsub/IDataSubscriber.hpp"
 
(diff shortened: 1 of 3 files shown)


SilKit/include/silkit/detail/impl/services/pubsub/DataSubscriber.hpp
```diff
@@ -46,8 +47,8 @@ class DataSubscriber : public SilKit::Services::PubSub::IDataSubscriber
     inline void SetDataMessageHandler(SilKit::Services::PubSub::DataMessageHandler handler) override;
 
 private:
-    inline static void TheDataMessageHandler(void* context, SilKit_DataSubscriber* subscriber,
-                                             const SilKit_DataMessageEvent* dataMessageEvent);
+    inline static void SilKitCALL TheDataMessageHandler(void* context, SilKit_DataSubscriber* subscriber,
```
MariusBgm, line 50: I searched for other instances, but these three handlers are the only ones having a proper static method. I suppose the only reason is because it's used in the Ctor.

MariusBgm (approved):
LGTM. couldn't find other instances of this pattern, too bad we can't use lambdas here

SilKit/include/silkit/detail/impl/services/pubsub/DataSubscriber.hpp
VDanielEdwards: I think we used the static methods because the same handler is used in two places (constructor and the `Set...Handler` method), and we wanted to avoid duplicating the lambda.

We could wrap the call to `The...Handler` in a lambda.

MariusBgm: The only advantage would be that we don't have to pull the SilKitMacros into this header. I think we can stick to the methods. at least, they are straight forward

PR #196, opened 2025-04-07
hop-on: send current sim. task instead of next inside of sim. step

VDanielEdwards:
## Subject

This change fixes an issue with hopping onto running time-synchronized simulations with a participant with an autonomous lifecycle.

The bug can cause a participant to release it's sim step 'early' and cause other participants to execute their next simulation step too early.

**TODO**:
 - [x] Rework the log messages

- [ ] Changelog.md updated

### DIFF
--- a/SilKit/source/services/orchestration/TimeSyncService.cpp
+++ b/SilKit/source/services/orchestration/TimeSyncService.cpp
@@ -370,7 +370,17 @@ TimeSyncService::TimeSyncService(Core::IParticipantInternal* participant, ITimeP
                                   "Participant \'{}\' is joining an already running simulation. Resending our "
                                   "NextSimTask.",
                                   descriptorParticipantName);
-                            SendMsg(_timeConfiguration.NextSimStep());
+
+                            if (GetTimeSyncPolicy()->IsExecutingSimStep())
+                            {
+                                Debug(_participant->GetLogger(), "Sending currently executing simulation step");
+                                SendMsg(_timeConfiguration.CurrentSimStep());
+                            }
+                            else
+                            {
+                                Debug(_participant->GetLogger(), "Sending next simulation step");
+                                SendMsg(_timeConfiguration.NextSimStep());
+                            }
                         }
                     }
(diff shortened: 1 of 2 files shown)


SilKit/source/services/orchestration/TimeSyncService.cpp
```diff
@@ -370,7 +370,17 @@ TimeSyncService::TimeSyncService(Core::IParticipantInternal* participant, ITimeP
                                   "Participant \'{}\' is joining an already running simulation. Resending our "
                                   "NextSimTask.",
                                   descriptorParticipantName);
-                            SendMsg(_timeConfiguration.NextSimStep());
+
+                            if (GetTimeSyncPolicy()->IsExecutingSimStep())
```
MariusBgm, line 374: I wonder how we could test this with an integration test. we would have to add some nested participant, that requests nextSimTask while another one is hanging in SimulationStepAsync. Would that be feasible?

MariusBgm (approved):
LGTM! however we should maybe add a test for this edge case

SilKit/source/services/orchestration/TimeSyncService.cpp
VDanielEdwards: The problematic behavior occurs when a participant performs a hop-on while the simulation step of another participant is still executing. Problem is that we would need to see all the `NextSimTask` messages, since under 'normal circumstances' – i.e., not having nested participants with weird synchronization between them – the system won't stop.

VDanielEdwards: We might be able to put it into a unit-test though. I'll have a look.

VDanielEdwards: Added two unit tests that ensure that the current (when hop-on happends during a simulation step) or the next (when hop-on happends outside of a simulation step) timestamp is used for the 'informative' message to the participant that is 'hopping on'.

MariusBgm: Thanks! looks good

PR #205, opened 2025-05-14
Heads-up: we are sunsetting distribution of Ubuntu 18.04 binary packages

MariusBgm:
Ubuntu 18.04 is EOL, and we should provide up-to-date Ubuntu binary packages.


VDanielEdwards (approved):
LGTM! 👍 Good riddance to 18.04!

PR #209, opened 2025-05-19
update changelog for 2025 QA2 release

MariusBgm:
We forgot to fix the Changelog date for the last RC


VJanKraemer (approved):
date seems correct, so lgtm

PR #199, opened 2025-04-15
SILKIT-1379: fix mocks and change metric kind serialization

NadegeNag:
## Developer checklist (address before review)

- [ ] Changelog.md updated
- [ ] Prepared update for depending repositories
- [ ] Documentation updated (public API changes only)
- [ ] API docstrings updated (public API changes only)
- [ ] Rebase &rarr; commit history clean
- [ ] Squash and merge &rarr; proper PR title

PR #215, opened 2025-06-06
add metrics to dashboard

MariusBgm:
This PR extends the work done in #92 and #199.
Outline:
- add "Attribute" metric, which is constant
- add metrics for bytes transferred, transmit queue size and messages transfered
- make the metric name a list of strings, or as fallback, split a single string using the `/` separator into a list

TODO: automated testing proved difficult, I'm still investigating.

### DIFF
--- a/SilKit/source/dashboard/Dto/BulkUpdateDto.hpp
+++ b/SilKit/source/dashboard/Dto/BulkUpdateDto.hpp
@@ -127,6 +127,12 @@ class BulkSimulationDto : public oatpp::DTO
     }
 };
 
+ENUM(MetricKind, v_int32, //
+     VALUE(Unknown, 0, "unknown"), //
+     VALUE(Counter, 10, "counter"), //
+     VALUE(Statistic, 20, "statistic"), //
+     VALUE(StringList, 30, "stringlist"))
+
 class MetricDataDto : public oatpp::DTO
 {
     DTO_INIT(MetricDataDto, DTO)
@@ -134,7 +140,7 @@ class MetricDataDto : public oatpp::DTO
     DTO_FIELD(Int64, ts);
     DTO_FIELD(String, pn);
     DTO_FIELD(String, mn);
-    DTO_FIELD(String, mk);
+    DTO_FIELD(Enum<MetricKind>::AsString, mk);
     DTO_FIELD(String, mv);
 };
 
--- a/SilKit/source/dashboard/Dto/BulkUpdateDto.hpp
+++ b/SilKit/source/dashboard/Dto/BulkUpdateDto.hpp
@@ -127,6 +127,31 @@ class BulkSimulationDto : public oatpp::DTO
     }
 };
 
+class MetricDataDto : public oatpp::DTO
+{
+    DTO_INIT(MetricDataDto, DTO)
+
+    DTO_FIELD(Int64, ts);
+    DTO_FIELD(String, pn);
+    DTO_FIELD(String, mn);
+    DTO_FIELD(String, mk);
+    DTO_FIELD(String, mv);
+};
+
+class MetricsUpdateDto : public oatpp::DTO
+{
+    DTO_INIT(MetricsUpdateDto, DTO)
+
+    DTO_FIELD(Vector<Object<MetricDataDto>>, metrics);
+
+    static auto CreateEmpty() -> Wrapper
+    {
+        auto dto = createShared();
+        dto->metrics = Vector<Object<MetricDataDto>>::createShared();
+        return dto;
+    }
+};
+
 } // namespace Dashboard
 } // namespace SilKit
 
--- a/SilKit/source/dashboard/Dto/BulkUpdateDto.hpp
+++ b/SilKit/source/dashboard/Dto/BulkUpdateDto.hpp
@@ -66,42 +66,26 @@ class BulkParticipantDto : public oatpp::DTO
     DTO_INIT(BulkParticipantDto, DTO)
 
     DTO_FIELD(String, name);
-    DTO_FIELD(Vector<Object<ParticipantStatusDto>>, statuses);
-    DTO_FIELD(Vector<Object<BulkControllerDto>>, canControllers);
-    DTO_FIELD(Vector<Object<BulkControllerDto>>, ethernetControllers);
-    DTO_FIELD(Vector<Object<BulkControllerDto>>, flexrayControllers);
-    DTO_FIELD(Vector<Object<BulkControllerDto>>, linControllers);
-    DTO_FIELD(Vector<Object<BulkDataServiceDto>>, dataPublishers);
-    DTO_FIELD(Vector<Object<BulkDataServiceDto>>, dataSubscribers);
-    DTO_FIELD(Vector<Object<BulkServiceInternalDto>>, dataSubscriberInternals);
-    DTO_FIELD(Vector<Object<BulkRpcServiceDto>>, rpcClients);
-    DTO_FIELD(Vector<Object<BulkRpcServiceDto>>, rpcServers);
-    DTO_FIELD(Vector<Object<BulkServiceInternalDto>>, rpcServerInternals);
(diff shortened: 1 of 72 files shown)


SilKit/source/services/metrics/MetricsDatatypes.hpp
```diff
@@ -32,6 +33,7 @@ struct MetricData
     std::string name;
     MetricKind kind;
     std::string value;
+    std::vector<std::string> nameList; // supersedes name
```
nadegegriesser: as it is an experimental internal api, we can change the type of name instead of introducing a new field

SilKit/source/services/metrics/MetricsDatatypes.hpp
```diff
@@ -23,6 +23,7 @@ enum struct MetricKind
     COUNTER,
     STATISTIC,
     STRING_LIST,
```
nadegegriesser, line 27: do we still have "metrics" of type stringlist? otherwise it can be removed

SilKit/source/dashboard/Dto/BulkUpdateDto.hpp
```diff
@@ -127,6 +127,38 @@ class BulkSimulationDto : public oatpp::DTO
     }
 };
 
+ENUM(MetricKind, v_int32,
+     VALUE(Unknown, 0, "unknown"),
```
nadegegriesser: is unknown needed?

SilKit/source/dashboard/Dto/BulkUpdateDto.hpp
```diff
@@ -127,6 +127,38 @@ class BulkSimulationDto : public oatpp::DTO
...
+    DTO_FIELD(String, pn);
+    DTO_FIELD(Vector<String>, mn);
+    DTO_FIELD(Enum<MetricKind>::AsString, mk);
+    DTO_FIELD(String, mv);
+};
+
+class MetricsUpdateDto : public oatpp::DTO
+{
+    DTO_INIT(MetricsUpdateDto, DTO)
+
```
nadegegriesser, line 134: maybe we could split the metrics by type, it would make type validation easier and decrease the payload size, for example:

remove MetricKind an make following changes:

`class MetricDataDto: public oatpp::DTO`
`{`
`   DTO_INIT(MetricDataDto, DTO)`
` `
`    DTO_FIELD(Int64, ts);`
`    DTO_FIELD(String, pn);`
`    DTO_FIELD(Vector<String>, mn);`
`};` 
` `
`class AttributeDataDto: public MetricDataDto`
`{`
`   DTO_INIT(AttributeDataDto, MetricDataDto)`
` `
`    DTO_FIELD(String, mv);`
`};` 
` `
`class CounterDataDto: public MetricDataDto`
`{`
`   DTO_INIT(CounterDataDto, MetricDataDto)`
` `
`    DTO_FIELD(Int64, mv);`
`};` 
` `
`class StatisticDataDto: public MetricDataDto`
`{`
`   DTO_INIT(StatisticDataDto, MetricDataDto)`
` `
`    DTO_FIELD(Vector<Float64>, mv);`
`};` 
` `
`class MetricsUpdateDto : public oatpp::DTO`
`{`
`   DTO_INIT(MetricsUpdateDto, oatpp::DTO)`
` `
`    DTO_FIELD(Vector<Object<AttributeDataDto>>, attributes);`
`    DTO_FIELD(Vector<Object<CounterDataDto>>, counters);`
`    DTO_FIELD(Vector<Object<StatisticDataDto>>, statistics);`
`};`

SilKit/source/services/metrics/IMetricsManager.hpp
```diff
@@ -5,20 +5,44 @@
...
+
+
     virtual ~IMetricsManager() = default;
     virtual void SubmitUpdates() = 0;
-    virtual auto GetCounter(const std::string& name) -> ICounterMetric* = 0;
-    virtual auto GetStatistic(const std::string& name) -> IStatisticMetric* = 0;
-    virtual auto GetStringList(const std::string& name) -> IStringListMetric* = 0;
+    virtual auto GetCounter(MetricName name) -> ICounterMetric* = 0;
+    virtual auto GetStatistic(MetricName name) -> IStatisticMetric* = 0;
+    virtual auto GetStringList(MetricName name) -> IStringListMetric* = 0;
```
nadegegriesser, line 23: maybe not needed

SilKit/source/dashboard/Dto/BulkUpdateDto.hpp
```diff
@@ -127,6 +127,38 @@ class BulkSimulationDto : public oatpp::DTO
     }
 };
 
+ENUM(MetricKind, v_int32,
+     VALUE(Unknown, 0, "unknown"),
+     VALUE(Counter, 10, "counter"),
+     VALUE(Statistic, 20, "statistic"),
+     VALUE(StringList, 30, "stringlist"),
```
nadegegriesser: is stringlist needed?

SilKit/source/services/metrics/MetricsDatatypes.hpp
MariusBgm: ok, will do

SilKit/source/dashboard/Dto/BulkUpdateDto.hpp
MariusBgm: Will attempt this

SilKit/source/dashboard/Dto/BulkUpdateDto.hpp
MariusBgm: it is used for the `TcpAcceptors` and `LocalAcceptors` of a participant. I could reuse the Attribute for it, but would have to encode its values as a JSON list... Maybe i could just rename StringList to Attribute?

SilKit/source/services/metrics/IMetricsManager.hpp
```diff
@@ -5,20 +5,42 @@
...
 namespace VSilKit {
 
 struct ICounterMetric;
 struct IStatisticMetric;
 struct IStringListMetric;
+struct IAttributeMetric;
+
+using MetricName = std::initializer_list<std::string>;
+
+inline auto ToString(MetricName stringList) -> std::string
```
VDanielEdwards: This function always produces strings that end with '/'. Is this correct?

I'd propose changing the implementation to

```c++
    if (stringList.size() == 0)
    {
        return {};
    }

    auto it = stringList.begin();

    std::stringstream ss;
    ss << *it;

    for (++it; it != stringList.end(); ++it)
    {
        ss << '/' << *it;
    }

    return ss.str();
```

and moving the implementation into a TU.

Since this will also always copy the strings into a heap-allocation when used with 'naked' `"abc"` string literals, we could consider using `std::initializer_list<std::string_view>` here instead of `std::string`.

SilKit/source/util/StringHelpers.cpp
```diff
@@ -141,5 +141,20 @@ auto PrintableString(const std::string& participantName) -> std::string
     return safeName;
 }
 
+auto SplitString(std::string input, const std::string& separator)  -> std::vector<std::string>
```
VDanielEdwards: If we replace all `std::string` in the arguments with `std::string_view` this method would only allocate the result strings. When `std::string` is used it will also copy all the intermediate `[i+1..]` substrings.

VDanielEdwards (approved):
LGTM! 👍 Found some minor things, but nothing huge.

SilKit/source/util/StringHelpers.cpp
MariusBgm: omg, we have string_view now. will do! :-)

SilKit/source/services/metrics/IMetricsManager.hpp
MariusBgm: will do

MariusBgm: I moved the MetricName to MetricsDataTypes.hpp/cpp  for now

PR #218, opened 2025-06-25
config: add export final participant configuration as string

MariusBgm:
When a participant configuration was processed, it may include fragments included by `includes:` statements.
This adds a new API to retrieve the finalized participant configuration as string.
Will be used e.g. in the network simulator and the dashboard.

### DIFF
--- a/SilKit/include/silkit/capi/SilKit.h
+++ b/SilKit/include/silkit/capi/SilKit.h
@@ -104,4 +104,14 @@ typedef SilKit_ReturnCode(SilKitFPTR* SilKit_ParticipantConfiguration_Destroy_t)
     SilKit_ParticipantConfiguration* participantConfiguration);
 
 
+/*! \brief Return a JSON-string of the complete parsed participant configuration.
+*  \param outputJsonString the JSON string of the configuration. When this is NULL outputSize will contain the required size for the JSON string.
+*  \param outputSize the size of the outputJsonString is stored here. Can be called with outputJsonString set to NULL to get the required size.
+*/
+SilKitAPI SilKit_ReturnCode SilKitCALL SilKit_ParticipantConfiguration_ToString(
+    const SilKit_ParticipantConfiguration* outParticipantConfiguration, char** outputJsonString, size_t* outputSize);
+
+typedef SilKit_ReturnCode(SilKitFPTR* SilKit_ParticipantConfiguration_ToString_t)(
+    const SilKit_ParticipantConfiguration* outParticipantConfiguration, char** outputJsonString, size_t* outputSize);
+
 SILKIT_END_DECLS
--- a/SilKit/include/silkit/config/IParticipantConfiguration.hpp
+++ b/SilKit/include/silkit/config/IParticipantConfiguration.hpp
@@ -73,6 +73,9 @@ DETAIL_SILKIT_CPP_API auto ParticipantConfigurationFromString(const std::string&
 DETAIL_SILKIT_CPP_API auto ParticipantConfigurationFromFile(const std::string& filename)
     -> std::shared_ptr<SilKit::Config::IParticipantConfiguration>;
 
+DETAIL_SILKIT_CPP_API auto ParticipantConfigurationToString(
+    std::shared_ptr<SilKit::Config::IParticipantConfiguration> config) -> std::string;
+
 } // namespace Config
 DETAIL_SILKIT_DETAIL_VN_NAMESPACE_CLOSE
 } // namespace SilKit
--- a/SilKit/include/silkit/detail/impl/config/IParticipantConfiguration.ipp
+++ b/SilKit/include/silkit/detail/impl/config/IParticipantConfiguration.ipp
@@ -26,6 +26,9 @@
 #include "silkit/detail/impl/ThrowOnError.hpp"
 #include "silkit/detail/impl/config/ParticipantConfiguration.hpp"
 
+#include <string>
+#include <vector>
+
 namespace SilKit {
 DETAIL_SILKIT_DETAIL_VN_NAMESPACE_BEGIN
 namespace Config {
@@ -52,6 +55,25 @@ auto ParticipantConfigurationFromFile(const std::string& path)
     return std::make_shared<Impl::Config::ParticipantConfiguration>(participantConfiguration);
 }
 
+auto ParticipantConfigurationToString(std::shared_ptr<SilKit::Config::IParticipantConfiguration> config) -> std::string
+{
+    size_t size{};
+    auto&& concreteConfig = std::dynamic_pointer_cast<Impl::Config::ParticipantConfiguration>(config);
+    SilKit_ParticipantConfiguration_ToString(concreteConfig->Get(), nullptr, &size);
+
+    std::vector<char> buffer;
+    if( size > 0)
+    {
+        buffer.resize(size);
+        //C++17 std::string::data() -> char*;
+        auto&& data = buffer.data();
+        SilKit_ParticipantConfiguration_ToString(concreteConfig->Get(), &data, &size);
+
+    }
+    return {buffer.data(), buffer.size()};
+}
+
+
 } // namespace Config
 DETAIL_SILKIT_DETAIL_VN_NAMESPACE_CLOSE
 } // namespace SilKit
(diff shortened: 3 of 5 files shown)


SilKit/include/silkit/capi/SilKit.h
```diff
@@ -104,4 +104,14 @@ typedef SilKit_ReturnCode(SilKitFPTR* SilKit_ParticipantConfiguration_Destroy_t)
     SilKit_ParticipantConfiguration* participantConfiguration);
 
 
+/*! \brief Return a JSON-string of the complete parsed participant configuration.
+*  \param outputJsonString the JSON string of the configuration. When this is NULL outputSize will contain the required size for the JSON string.
+*  \param outputSize the size of the outputJsonString is stored here. Can be called with outputJsonString set to NULL to get the required size.
+*/
+SilKitAPI SilKit_ReturnCode SilKitCALL SilKit_ParticipantConfiguration_ToString(
+    const SilKit_ParticipantConfiguration* outParticipantConfiguration, char** outputJsonString, size_t* outputSize);
```
VDanielEdwards: The C-API doesn't really care about the const-ness of the pointers. I'd keep it simple and just use a non-const one for the 'this' argument. As it's not an out-parameter, the name shouldn't be prefixed with `out`.

```suggestion
SilKitAPI SilKit_ReturnCode SilKitCALL SilKit_ParticipantConfiguration_ToString(
    SilKit_ParticipantConfiguration* participantConfiguration, char** outputJsonString, size_t* outputSize);
```

SilKit/include/silkit/capi/SilKit.h
```diff
@@ -104,4 +104,14 @@ typedef SilKit_ReturnCode(SilKitFPTR* SilKit_ParticipantConfiguration_Destroy_t)
...
+
+typedef SilKit_ReturnCode(SilKitFPTR* SilKit_ParticipantConfiguration_ToString_t)(
+    const SilKit_ParticipantConfiguration* outParticipantConfiguration, char** outputJsonString, size_t* outputSize);
```
VDanielEdwards:
```suggestion
typedef SilKit_ReturnCode(SilKitFPTR* SilKit_ParticipantConfiguration_ToString_t)(
    SilKit_ParticipantConfiguration* participantConfiguration, char** outputJsonString, size_t* outputSize);
```

SilKit/include/silkit/capi/SilKit.h
```diff
@@ -104,4 +104,14 @@ typedef SilKit_ReturnCode(SilKitFPTR* SilKit_ParticipantConfiguration_Destroy_t)
     SilKit_ParticipantConfiguration* participantConfiguration);
 
 
+/*! \brief Return a JSON-string of the complete parsed participant configuration.
```
VDanielEdwards, line 90: I think if we want to nail the format down to JSON, we can call the function `ToJson`.

SilKit/include/silkit/config/IParticipantConfiguration.hpp
```diff
@@ -73,6 +73,9 @@ DETAIL_SILKIT_CPP_API auto ParticipantConfigurationFromString(const std::string&
 DETAIL_SILKIT_CPP_API auto ParticipantConfigurationFromFile(const std::string& filename)
     -> std::shared_ptr<SilKit::Config::IParticipantConfiguration>;
 
+DETAIL_SILKIT_CPP_API auto ParticipantConfigurationToString(
```
VDanielEdwards: We could also add this to the C++ interface as it's first method.

SilKit/include/silkit/detail/impl/config/IParticipantConfiguration.ipp
```diff
@@ -52,6 +55,25 @@ auto ParticipantConfigurationFromFile(const std::string& path)
     return std::make_shared<Impl::Config::ParticipantConfiguration>(participantConfiguration);
 }
 
+auto ParticipantConfigurationToString(std::shared_ptr<SilKit::Config::IParticipantConfiguration> config) -> std::string
+{
+    size_t size{};
+    auto&& concreteConfig = std::dynamic_pointer_cast<Impl::Config::ParticipantConfiguration>(config);
+    SilKit_ParticipantConfiguration_ToString(concreteConfig->Get(), nullptr, &size);
```
VDanielEdwards: Should check the return code here. Otherwise `size` might be invalid.

SilKit/include/silkit/detail/impl/config/IParticipantConfiguration.ipp
```diff
@@ -52,6 +55,25 @@ auto ParticipantConfigurationFromFile(const std::string& path)
...
+
+    std::vector<char> buffer;
+    if( size > 0)
+    {
+        buffer.resize(size);
+        //C++17 std::string::data() -> char*;
+        auto&& data = buffer.data();
+        SilKit_ParticipantConfiguration_ToString(concreteConfig->Get(), &data, &size);
```
VDanielEdwards: This line should also check the return code. Otherwise `data` might be invalid.

SilKit/source/capi/CapiParticipant.cpp
```diff
@@ -155,6 +156,29 @@ try
...
+{
+    ASSERT_VALID_OUT_PARAMETER(participantConfiguration);
+    ASSERT_VALID_POINTER_PARAMETER(requiredSize);
+    //outputString may be NULL
+
+    auto* cppParticipantConfiguration =
+        reinterpret_cast<const SilKit::Config::ParticipantConfiguration*>(participantConfiguration);
+
+    auto&& jsonString = SilKit::Config::SerializeAsJson(*cppParticipantConfiguration);
+    *requiredSize = jsonString.size();
```
VDanielEdwards, line 155: We need to keep the value behind `requiredSize` around, since if `outputString != nullptr` this gives us the size of the buffer.

SilKit/source/capi/CapiParticipant.cpp
```diff
@@ -155,6 +156,29 @@ try
...
+
+    if(outputString != nullptr && jsonString.size() > 0)
+    {
+        std::copy(std::cbegin(jsonString), std::cend(jsonString), *outputString);
```
VDanielEdwards: I wouldn't check the size, but just copy `min(*requiredSize, jsonString.size())` characters. Not sure if we should write the `\0` at the end.

VDanielEdwards (changes requested):

SilKit/source/capi/CapiParticipant.cpp
VDanielEdwards: Alternatively return an error code if the buffer is too small.

VDanielEdwards: Not writing the `\0` makes it a bit more friendly for API users that don't use nul-terminated strings.

SilKit/include/silkit/capi/SilKit.h
MariusBgm: I really like to have const as part of the API as a way of documentation. we really don't modify this object -- the naming of the variable is definitely wrong, though

VDanielEdwards: Agreed, we can show it via the signature that way.

SilKit/IntegrationTests/ITest_Config.cpp
```diff
@@ -0,0 +1,92 @@
...
+    ASSERT_EQ(result, SilKit_ReturnCode_SUCCESS);
+
+    std::array<char, 1024> buf{};
+    std::memset(buf.data(), 'A', buf.size());
+    size_t size{};
+
+    // get size
+    result = SilKit_ParticipantConfiguration_ToJson(config, nullptr, &size);
+    ASSERT_EQ(result, SilKit_ReturnCode_SUCCESS);
+    ASSERT_GT(size, 0) << "the string should have length > 0";
```
VDanielEdwards, line 78: Just to avoid strange crashes due to buffer overflows in this test (if someone adjusts the yaml string). `_LT` to leave one character to check for 'no null terminator'.

```suggestion
    ASSERT_GT(size, 0) << "the string should have length > 0";
    ASSERT_LT(size, buf.size()) << "the string should have length <= buf.size()";
```

SilKit/include/silkit/detail/impl/config/IParticipantConfiguration.ipp
```diff
@@ -52,6 +38,26 @@ auto ParticipantConfigurationFromFile(const std::string& path)
...
+    size_t size{};
+    auto&& concreteConfig = std::dynamic_pointer_cast<Impl::Config::ParticipantConfiguration>(config);
+    auto returnCode = SilKit_ParticipantConfiguration_ToJson(concreteConfig->Get(), nullptr, &size);
+    Impl::ThrowOnError(returnCode);
+
+    std::vector<char> buffer;
+    if( size > 0)
+    {
+        buffer.resize(size);
+        //C++17 std::string::data() -> char*;
```
VDanielEdwards: `buffer` is a `std::vector`, the string is constructed later.
```suggestion
```

VDanielEdwards (approved):
LGTM! 👍 Some small suggestions, but nothing critical.

PR #221, opened 2025-06-26
config: fix unused parameter warnings and some styling

VDanielEdwards:
## Subject

Nightly build broke due to warnings (as errors). This should fix that.

- [ ] Changelog.md updated

### DIFF
--- a/SilKit/source/config/YamlReader.hpp
+++ b/SilKit/source/config/YamlReader.hpp
@@ -17,11 +17,13 @@
 namespace VSilKit {
 
 template <typename Impl>
-struct BasicYamlReader
+class BasicYamlReader
 {
+protected:
     ryml::Parser& _parser;
     ryml::ConstNodeRef _node;
 
+public:
     BasicYamlReader(ryml::Parser& parser_, ryml::ConstNodeRef node_)
         : _parser(parser_)
         , _node(node_)


SilKit/source/config/YamlWriter.hpp
```diff
@@ -14,102 +14,128 @@
...
-{
-    node |= ryml::MAP;
-}
 
-struct YamlWriter
+template <typename Impl>
+struct BasicYamlWriter
 {
-    ryml::NodeRef _node;
+    ryml::NodeRef node;
```
MariusBgm, line 21: actually, we should adhere to our style of naming it `_node`. Also, the YamlReader was merged with its YamlReaderContext and thus uses `node` throughout, now. we should stream line it again.

SilKit/source/config/YamlReader.hpp
```diff
@@ -15,40 +16,19 @@
...
 
-//Utilities
-bool IsValidChild(const ryml::ConstNodeRef& node, const std::string& name);
-
-struct YamlReaderContext
+template <typename Impl>
+struct BasicYamlReader
 {
     ryml::Parser& parser;
     ryml::ConstNodeRef node;
```
MariusBgm: This used to be a POD helper struct `YamlReaderContext`, so I neglected the style convention. should be `_parser` and `_node` for a class.

MariusBgm (approved):
LGTM! maybe adjust some member names according to convention. 

In the future, we should move the config parsing into an isolated, separate compilation unit. so we can consume it without SIL Kit internals.

Also, i saw alot of formatting changes: we should enforce clang-format on all new PRs, so the diffs get shorter in the future.

SilKit/source/config/YamlWriter.hpp
VDanielEdwards: The `BasicYamlReader` is now a `class` and has the fields `protected`. The field names are now prefixed with `_`.

PR #232, opened 2025-07-24
dashboard: anotate types to metric names

MariusBgm:
add explicit types in the metric names, and change from counter (ever increasing) to statistics.
also add bandwidth metrics.

SILKIT-1799


VDanielEdwards (approved):
LGTM! 👍

PR #233, opened 2025-07-24
Switch to relwithdebinfo

VJanKraemer:
## Subject
Suppor release builds with debuginfos

## Description
We added release builds with debuginfo added in our releases. This supports splitting the library from the debug symbols so we can have an extra symbol file that is installed.

- [ ] Changelog.md updated


SilKit/cmake/SilKitBuildTools.cmake
```diff
@@ -42,14 +41,12 @@ macro(silkit_package_debugsymbols targetName)
...
-        add_custom_command(
-            TARGET "${targetName}"
-            POST_BUILD
-            WORKING_DIRECTORY "${SILKIT_SYMBOLS_DIR_BASE}"
-            COMMAND "${CMAKE_COMMAND}" -E tar cf "${SILKIT_SYMBOLS_DIR_NAME}.zip" --format=zip -- "${SILKIT_SYMBOLS_DIR_NAME}"
-            COMMENT "SIL Kit: SILKIT_PACKAGE_SYMBOLS: creating ELF debug symbols zip"
-        )
+	message(STATUS "Installing ELF debug file")
+	install(
+		FILES "${SILKIT_SYMBOLS_DIR}/${targetName}${CMAKE_DEBUG_POSTFIX}.debug"
```
MariusBgm, line 46: So, we bundle the debug symbols as part of the CPack zip files? does this interfere with the debian builds?

MariusBgm (approved):
LGTM!

SilKit/cmake/SilKitBuildTools.cmake
VJanKraemer: No since we do not use any CPack stuff for Debian. That is packaged all via vanilla rules/control files in vectorgrp/sil-kit-pkg

PR #238, opened 2025-07-29
abi: remove legacy symbols 

MariusBgm:
When switchting to the hour-glass implementation (c++ frontend for the C-ABI) we provided some legacy ABI symbols of the old c++ implementation.
Remove this for the 5.0 release


VDanielEdwards (approved):
Good riddance! 👍

PR #242, opened 2025-07-30
qnx: remove c++14 from compile flags

VJanKraemer:
## Description
We are now using the C++17 standard so we can remove the c++14 flag

- [ ] Changelog.md updated

### DIFF
--- a/SilKit/cmake/qnx-cross-base-toolchain.cmake
+++ b/SilKit/cmake/qnx-cross-base-toolchain.cmake
@@ -69,7 +69,7 @@ set(CMAKE_CXX_COMPILER_TARGET ${qcc_arch})
 
 # Use LLVM stdlib for now, since GNU is segfaulting with future.waits
 # -Y and -stdlib should be redundant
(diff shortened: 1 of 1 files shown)


SilKit/cmake/qnx-cross-base-toolchain.cmake
```diff
@@ -69,7 +69,7 @@ set(CMAKE_CXX_COMPILER_TARGET ${qcc_arch})
 
 # Use LLVM stdlib for now, since GNU is segfaulting with future.waits
 # -Y and -stdlib should be redundant
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Y_cxx -stdlib=libc++ -Wc,-std=c++14")
```
MariusBgm, line 72: i forgot why i even put it there.
Does it compile cleanly on QNX 8 ?
not sure if we have to use the LLVM libc++ instead of the GNU libstdc++ on QNX8, on QNX7 it fails to build...

VJanKraemer:
Closed in favor of https://github.com/vectorgrp/sil-kit/pull/243

PR #244, opened 2025-07-30
cmake: add distrib preset for building release packages

MariusBgm:
have all cmake options needed for distributions packages bundled in the `distrib` preset. 
While at it, remove all the unnecessary presets, we now have:
```
$ cmake --list-presets
Available configure presets:

  "distrib"
  "debug"
  "release"
  "relwithdebinfo"
  "x86-debug"
  "x86-release"
  "x86-relwithdebinfo"
  "vs141-x86-debug"
  "vs141-x64-debug"
  "cross-armeabihf-debug"
  "cross-armeabi-debug"
  "cross-qnx-x86_64"      - QNX cross-build x86_64
  "cross-qnx-armv7"       - QNX cross-build armv7
  "cross-qnx-aarch64"     - QNX cross-build aarch64
```


.github/workflows/linux-ubsan.yml
```diff
@@ -22,8 +22,9 @@ jobs:
...
       - uses: ./.github/actions/build-cmake-preset
         with:
-          preset-name: clang-relwithdebinfo
-          cmake-args: "-D SILKIT_ENABLE_UBSAN=ON -D SILKIT_BUILD_DASHBOARD=OFF \
+          preset-name: relwithdebinfo
+          cmake-args: "-D SILKIT_BUILD_DASHBOARD=OFF \
                        -DCMAKE_C_COMPILER=clang-18 \
-                       -DCMAKE_CXX_COMPILER=clang++-18"
+                       -DCMAKE_CXX_COMPILER=clang++-18 \
+                       -DCMAKE_CXX_FLAGS='-fsanitize=undefined -fno-omit-frame-pointer -g3'"
```
VDanielEdwards: The `-g3` flags will be overwritten by the CMake provided flags, as the value of `CMAKE_CXX_FLAGS` is _prepended_ to the build-type dependent flags.

Same goes for the other workflows.

We can however simply overwrite those by hand:

```suggestion
                       -DCMAKE_CXX_FLAGS='-fsanitize=undefined -fno-omit-frame-pointer'" \
                       -DCMAKE_CXX_FLAGS_RELWITHDEBINFO='-Og -g3'"
```

VDanielEdwards: The default for RelWithDebInfo is `-O2 -g` btw. And later occurences on the `gcc` or `clang` command line override the earlier ones.

VDanielEdwards: We don't have to use `-Og` btw., just as an example here.

VJanKraemer: Do you need to set CMAKE_CXX_FLAGS_<CONFIG> or CMAKE_CXX_FLAGS_<CONFIG>_INIT  🤔

VJanKraemer: The none `_INIT` version, so forget the previous comment

MariusBgm: i like using `-Og` as it adds some debugging optimizations. and using the `CMAKE_CXX_FLAGS_${CMAKE_BUILD_TYPE}` is good enough for the CI.

VDanielEdwards: From my own experiments, the `_INIT` values got overwritten somewhere somehow.

.github/workflows/linux-asan.yml
```diff
@@ -21,8 +21,10 @@ jobs:
...
         with:
-          preset-name: clang-relwithdebinfo
-          cmake-args: "-D SILKIT_ENABLE_ASAN=ON -D SILKIT_BUILD_DASHBOARD=OFF \
+          preset-name: relwithdebinfo
+          cmake-args: "-D SILKIT_BUILD_DASHBOARD=OFF \
                        -DCMAKE_C_COMPILER=clang-18 \
-                       -DCMAKE_CXX_COMPILER=clang++-18"
+                       -DCMAKE_CXX_COMPILER=clang++-18 \
+                       -DCMAKE_CXX_FLAGS='-fsanitize=address -fno-omit-frame-pointer' \
+                       -DCMAKE_CXX_FLAGS_RELWITHDEBINFO='-Og -g3'"
```
VJanKraemer, line 29: Do we need `-DNDEBUG` which I think is standard for Release and RelWithDebInfo.

MariusBgm: i think NDEBUG only disables cassert's `assert()` on GNU. we should be fine without setting it

VDanielEdwards: I don't see an issue adding it, nor with removing it. Since we don't use `assert`, the only place it might be used are our third-party libraries.

.github/workflows/linux-ubsan.yml
```diff
@@ -22,8 +22,10 @@ jobs:
...
         with:
-          preset-name: clang-relwithdebinfo
-          cmake-args: "-D SILKIT_ENABLE_UBSAN=ON -D SILKIT_BUILD_DASHBOARD=OFF \
+          preset-name: relwithdebinfo
+          cmake-args: "-D SILKIT_BUILD_DASHBOARD=OFF \
                        -DCMAKE_C_COMPILER=clang-18 \
-                       -DCMAKE_CXX_COMPILER=clang++-18"
+                       -DCMAKE_CXX_COMPILER=clang++-18 \
+                       -DCMAKE_CXX_FLAGS='-fsanitize=undefined -fno-omit-frame-pointer' \
+                       -DCMAKE_CXX_FLAGS_RELWITHDEBINFO='-Og -g3'"
```
VDanielEdwards, line 30: If the goal is to use `RelWithDebInfo` as the deployment build, we should probably benchmark if `-Og` hurts performance in contrast to `-O2` or `-O3`.

The `-Og` level doesn't do all optimizations in `-O1` which may be fine for us.

I'm all for using it in the debug build however.

VDanielEdwards: I'm generally happy if I get a halfway decent stacktrace from the deployment build - but I'm not overly salty about missing variables and stuff like that. I would for a debug build though ^^

VDanielEdwards: But since this ins in the `ubsan` build this is fine.

VJanKraemer (approved):
LGTM

PR #248, opened 2025-08-05
build: fix building Demos from source

MariusBgm:
This issue pops up, when building from the sources bundled with a distrib zip file, e.g. when you run:
```
cmake -S SilKit-Demos/ -B build-demos-debug -D CMAKE_BUILD_TYPE=Debug
```
the CMake bails out


VJanKraemer (approved):
LGTM

PR #259, opened 2025-09-08
log: fix json logging without keyvals

MariusBgm:
Logging messages without `kv`, that is a single `JsonString` was broken. The user-messages was not formatted as JSON.


VLukasBecker (approved):
lgtm!

PR #263, opened 2025-09-15
struct logs: add raw key for JSON blobs

MariusBgm:
The `raw` field in slogs allows us to add debugging information in JSON.
In this case, add the FR ControllerConfig as a JSON object.

SILKIT-1827

### DIFF
--- a/SilKit/source/services/logging/MessageTracing.hpp
+++ b/SilKit/source/services/logging/MessageTracing.hpp
@@ -15,21 +15,6 @@
 namespace SilKit {
 namespace Services {
 
-
-template <typename MsgT>
-std::chrono::nanoseconds GetTimestamp(MsgT& msg,
-                                      std::enable_if_t<Core::HasTimestamp<MsgT>::value, bool> = true)
-{
-    return msg.timestamp;
-}
-
-template <typename MsgT>
-std::chrono::nanoseconds GetTimestamp(MsgT& /*msg*/,
-                  std::enable_if_t<!Core::HasTimestamp<MsgT>::value, bool> = false)
-{
-    return std::chrono::nanoseconds::duration::min();
-}
-
 namespace Detail {
 template <class SilKitMessageT>
 void TraceMessageCommon(Logging::ILoggerInternal* logger,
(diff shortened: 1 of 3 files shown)


SilKit/source/services/logging/MessageTracing.hpp
```diff
@@ -29,8 +31,6 @@ std::chrono::nanoseconds GetTimestamp(MsgT& /*msg*/,
 }
 
 
-
-
 template <class SilKitMessageT>
```
KonradBreitsprecherBkd, line 19: We could reduce the code duplication here. 
With a `SilKitMsgTraitLogRaw` we also could get rid of the specialization on `FlexrayControllerConfig`.

```
namespace Detail {

template <class SilKitMessageT>
void TraceMessageCommon(Logging::ILoggerInternal* logger,
                       Logging::Level level,
                       const char* direction,
                       const Core::IServiceEndpoint* addr,
                       const SilKitMessageT& msg,
                       const Core::ServiceDescriptor* from = nullptr,
                       const std::string_view* target = nullptr)
{
    if (logger->GetLogLevel() == level)
    {
        Logging::LoggerMessage lm{logger, level};
        lm.SetMessage(direction);
        lm.SetKeyValue(addr->GetServiceDescriptor());
        lm.FormatKeyValue(Logging::Keys::msg, "{}", msg);

        if (from)
        {
            lm.SetKeyValue(Logging::Keys::from, from->GetParticipantName());
        }
        if (target)
        {
            lm.FormatKeyValue(Logging::Keys::to, "{}", *target);
        }

        auto virtualTimeStamp = GetTimestamp(msg);
        if (virtualTimeStamp != std::chrono::nanoseconds::duration::min())
        {
            lm.FormatKeyValue(Logging::Keys::virtualTimeNS, "{}", virtualTimeStamp.count());
        }
        if constexpr (std::is_same_v<SilKitMessageT, SilKit::Services::Flexray::FlexrayControllerConfig>)
        {
            lm.SetKeyValue(Logging::Keys::raw, SilKit::Config::SerializeAsJson(msg));
        }
        lm.Dispatch();
    }
}

} // namespace Detail

template <class SilKitMessageT>
void TraceRx(Logging::ILoggerInternal* logger, const Core::IServiceEndpoint* addr, const SilKitMessageT& msg,
             const Core::ServiceDescriptor& from)
{
    Detail::TraceMessageCommon(logger, Logging::Level::Trace, "Recv message", addr, msg, &from, nullptr);
}

template <class SilKitMessageT>
void TraceTx(Logging::ILoggerInternal* logger, const Core::IServiceEndpoint* addr, const SilKitMessageT& msg)
{
    Detail::TraceMessageCommon(logger, Logging::Level::Trace, "Send message", addr, msg, nullptr, nullptr);
}

template <class SilKitMessageT>
void TraceTx(Logging::ILoggerInternal* logger, const Core::IServiceEndpoint* addr, const std::string_view target, const SilKitMessageT& msg)
{
    Detail::TraceMessageCommon(logger, Logging::Level::Trace, "Send targetted message", addr, msg, nullptr, &target);
}
```

MariusBgm: thanks @KonradBreitsprecherBkd . took your proposal and tweaked it a bit more. i noted that incremental linking on my machine takes a very long time (for the old and the new templated trace messages).

MariusBgm: I also replaced the GetTimestamp with an inline `if constexpr (Core::HasTimestamp<SilKitMessageT>::value)`, and cleaned up the usage of `GetCurrentTime` so it doesn't collide with windows `#defines`

PR #265, opened 2025-09-19
Fix CMake configuration for QNX targets

HasanGrosic:
## Subject

 1. Fixes the wrong toolset selection for QNX aarch64
 2. Sets the correct default installation path for the QNX CMake presets

## Description

1. The QNX build for aarch64 was failing due to a wrong target toolset selection. When the "_cxx" suffix is appended to the originally selected "gcc_ntoaarch64le_gpp", it results in an unknown toolset target "gcc_ntoaarch64le_gpp_cxx". By removing the "_gpp" suffix, it results in a correct toolset target "gcc_ntoaarch64le_cxx".
2. The QNX presets did not have an `installDir` set, so the installation path was defaulting to `/usr/local`. Now the QNX presets inherit the `mixin-dirs` and have the same `installDir` as the other presets.

Are there any other files I missed with my fixes?

- [ ] Changelog.md updated

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -119,11 +119,11 @@
         {
             "name": "x-cross-qnx-x86_64",
             "generator": "Unix Makefiles",
+            "inherits": [ "mixin-dirs" ],
             "environment": {
                 "SILKIT_CMAKE_PRESET_BUILD_PLATFORM": "qnx"
             },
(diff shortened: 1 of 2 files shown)


VJanKraemer (approved):
LGTM. Thanks for providing the fixes :)

PR #266, opened 2025-09-23
add traces to history notifications

MariusBgm:
Add messages sent via the link's history to trace-logs.

### DIFF
--- a/SilKit/source/core/vasio/VAsioTransmitter.hpp
+++ b/SilKit/source/core/vasio/VAsioTransmitter.hpp
@@ -12,6 +12,8 @@
 #include "IMessageReceiver.hpp"
 #include "IServiceEndpoint.hpp"
 #include "traits/SilKitMsgTraits.hpp"
+#include "ILoggerInternal.hpp"
+#include "MessageTracing.hpp"
 
 #include "SerializedMessage.hpp"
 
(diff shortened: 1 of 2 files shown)


SilKit/source/core/vasio/VAsioTransmitter.hpp
```diff
@@ -49,12 +51,13 @@ struct MessageHistory<MsgT, 1>
...
         _hasValue = true;
     }
-    void NotifyPeer(IVAsioPeer* peer, EndpointId remoteIdx)
+    void NotifyPeer(Services::Logging::ILoggerInternal* logger, IVAsioPeer* peer, EndpointId remoteIdx)
     {
         if (!_hasValue || !_hasHistory)
             return;
 
         auto buffer = SerializedMessage(_last, _from, remoteIdx);
+        Services::TraceTx(logger, peer, _last);
```
KonradBreitsprecherBkd, line 60: Can we smuggle a small hint in `_last` that it's a  history message

MariusBgm: i think with the changes in https://github.com/vectorgrp/sil-kit/pull/263 we could add a structured log entry / raw object with some meta-infos.

PR #269, opened 2025-10-23
dco: Add DCO check for all contributions

VJanKraemer:
## Subject
Check whether commits are properly signed-off!

- [ ] Changelog.md updated

### DIFF
--- a/SilKit/ci/check_dco_signed.py
+++ b/SilKit/ci/check_dco_signed.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python3
+
+# SPDX-FileCopyrightText: 2025 Vector Informatik GmbH
+#
+# SPDX-License-Identifier: MIT
+
+import requests
+import argparse
+
+from ci_utils import die, warn
+
+
+def check_dco(url: str):
+    commits_url = url + '/commits'
+
+    r = requests.get(commits_url, verify=False)
+
+    all_commits_signed = True
+    for commitObject in r.json():
+        committer = commitObject["commit"]["committer"]
+        author = commitObject["commit"]["author"]
+        message = commitObject["commit"]["message"]
+
+        committer_error = ""
+        author_error = ""
+
+        # Do not check fixup messages
+        if "fixup" not in message:
+
+            # Check committer only if it they are different
+            if committer["email"] != author["email"] or committer["name"] != author["name"]:
+                sign_off = f'Signed-off-by: {committer["name"]} <{committer["email"]}>'
+                if sign_off not in message and committer["name"] != "GitHub":
+                    comitter_error = f'Signed-off-by from committer {committer["name"]} missing! Merging blocked!'
+
+            sign_off = f'Signed-off-by: {author["name"]} <{author["email"]}>'
+            if sign_off not in message:
+                author_error = f'Signed-off-by from author {author["name"]} missing, Merging Blocked!'
+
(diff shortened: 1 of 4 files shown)


SilKit/ci/check_dco_signed.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
+
+# SPDX-FileCopyrightText: 2024 Vector Informatik GmbH
+#
+# SPDX-License-Identifier: MIT
+
+import requests
```
MariusBgm, line 7: this seems to be implicitly installed. do we need to add this explicitly to our CI docker images?

VJanKraemer: So this runs on the regular Ubuntu images provided by Github (for now)
I wonder whether we should have a dedicated SIL Kit Python Image where we have a fully prepared python-venv ready, instead of installing everything globally via apt?

VDanielEdwards: Is there a reason to use the GItHub API at all? Due to the `checkout` step in the workflow job `check-dco-signed` the last commit of the `sil-kit` repository should be available.

Would it not be easier to parse the output of `git log -n 1 --pretty=%B` (which is the body of the last commit)? Also no need to disable any SSL verification in this case.

VDanielEdwards: The command `git log -1 --pretty="%H%n%an%n%ae%n%cn%n%ce%n%B"` gives me

- full commit hash
- author name
- author email
- committer name
- committer email
- complete body

of the last commit, separated by newlines. Is this different from what the GitHub API returns for a commit in a PR? Do they somehow fudge the committer?

```
db22fadfc1bafc2139e44651ad002312faed8d50
Daniel Edwards
daniel.edwards@vector.com
GitHub
noreply@github.com
capi: fix include cycle in EventProducer.h (#272)
```

VJanKraemer: But a PR can have multiple commits. I did not find another way to access all commits in the PR other than the PR API

VJanKraemer: git log target_branch..base_branch could work but we would still need to separate all commits by hand before checking

VDanielEdwards: When using `-z` the commits themselves will be separated with `NUL`. Example (with a local test branch, the `^@` are NUL characters):

```
$ git log -z --pretty="%H%n%an%n%ae%n%cn%n%ce%n%B" main..HEAD
dfc44ee4055e8bda203ab63966c78cc5f2facf25
Daniel Edwards
daniel.edwards@vector.com
Daniel Edwards
daniel.edwards@vector.com
much signoff

Signed-off-by: Daniel Edwards <daniel.edwards@vector.com>
^@31d525d71c69554882bc070b5cce17f0f268c540
Daniel Edwards
daniel.edwards@vector.com
Daniel Edwards
daniel.edwards@vector.com
such commit
^@bfa8a7392fb2d71a135c71be782060e497d2d020
Daniel Edwards
daniel.edwards@vector.com
Daniel Edwards
daniel.edwards@vector.com
much wow
^@
```

VDanielEdwards: If we want to stick with the GitHub API we could place a `requirements.txt` next to the script and create the venv 'on the fly'.

PR #273, opened 2025-10-30
Track connected participants to set workflow configuration late

KonradBreitsprecherBkd:
Currently the `sil-kit-system-controller` immediately propagates the required participants (via `SetWorkflowConfiguration`) on startup. This defines a system state which is calculated locally on each participant. If one of the requires participant drops out, the system state is `Error`. **Now, a restart of all required participant and the system controller is needed.** This behavior can be improved as the restart is actually not necessary.

With this PR, the connected/disconnected participants are tracked and only when all required participants have connected, `SetWorkflowConfiguration` is called. **This allows single participants to drop out and rejoin without having to restart other participants or the system controller**.

### DIFF
--- a/.github/workflows/sil-kit-ci.yml
+++ b/.github/workflows/sil-kit-ci.yml
@@ -46,6 +46,18 @@ jobs:
             python3 ./SilKit/ci/check_files_changed.py ${{ github.repository }} ${{ github.event.number }}; \
           fi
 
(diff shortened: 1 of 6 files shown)


VDanielEdwards (approved):
LGTM! 👍 It does make it much nicer to redo half of the setup, when you manually run a bunch of participants.

PR #274, opened 2025-10-31
core: allow for 0ns simulation steps and synchronization points

MariusBgm:
PoC: trigger a simulation step with 0ns duration, for certain message types (which implement the SIlKitMsgTrait<T>::IsSynchronizationPoint) 
TODO:
- need to guard the behavior with handshake capabilities? -> don't send zero timestamps to participants which do not understand them
- we should have an explicit synchronization message, independent of the NextSimTask. E.g., for flow control and backpressure in non-virtual-time participants TODO: look at barriers, e.g. the RequestReplyService.

### DIFF
--- a/SilKit/source/core/participant/Participant_impl.hpp
+++ b/SilKit/source/core/participant/Participant_impl.hpp
@@ -1327,7 +1327,7 @@ void Participant<SilKitConnectionT>::SendMsg(const IServiceEndpoint* from,
 
 template <typename SilKitConnectionT>
 template <typename MessageT>
-void Participant<SilKitConnectionT>::HandleSynchronizationPoint()
+void Participant<SilKitConnectionT>::HandleSynchronizationPoint(const IServiceEndpoint* service)
 {
     if constexpr (SilKitMsgTraits<RemoveCvRef<MessageT>>::IsSynchronizationPoint())
     {
@@ -1338,7 +1338,9 @@ void Participant<SilKitConnectionT>::HandleSynchronizationPoint()
             {
                 if(_participantConfig.enableSynchronizationPoints)
                 {
-                    timesync->TriggerSynchronization();
+                    const auto& serdesName = SilKitMsgTraits<RemoveCvRef<MessageT>>::SerdesName();
+                    const auto numReceivers = _connection.GetNumberOfRemoteReceivers(service, serdesName);
+                    timesync->TriggerSynchronization(numReceivers);
                 }
             }
         }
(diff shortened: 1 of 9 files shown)


SilKit/source/services/orchestration/TimeSyncService.cpp
```diff
@@ -112,39 +113,48 @@ struct SynchronizedPolicy : public ITimeSyncPolicy
 
     void ReceiveNextSimTask(const Core::IServiceEndpoint* from, const NextSimTask& task) override
     {
-        _configuration->OnReceiveNextSimStep(from->GetServiceDescriptor().GetParticipantName(), task);
+        if (_enableSynchronizationPoints && task == ZeroSimTask)
+        {
+            auto currentStep = _configuration->CurrentSimStep();
+            _controller.ExecuteSimStep(currentStep.timePoint, 0ns);
```
VDanielEdwards: We could just return-early and not indent the `switch` block below.

```suggestion
            _controller.ExecuteSimStep(currentStep.timePoint, 0ns);
            return;
```

SilKit/source/core/internal/traits/SilKitMsgTraits.hpp
```diff
@@ -14,6 +14,11 @@ namespace Core {
 // ==================================================================
 //  Trait which checks that '.timestamp' works
 // ==================================================================
+template<class T>
+struct RemoveCvRef
+{
+    using type = std::remove_cv_t<std::remove_reference_t<T>>;
+};
```
VDanielEdwards: You could switch to a templated typedef, then you don't have to use the `typename RemoveCvRef<...>::type` and can just write `RemoveCvRef<...>`.

```suggestion
template<class T>
using RemoveCvRef = std::remove_cv_t<std::remove_reference_t<T>>;
```

SilKit/source/services/orchestration/TimeConfiguration.cpp
```diff
@@ -77,7 +77,7 @@ void TimeConfiguration::OnReceiveNextSimStep(const std::string& participantName,
     {
         Logging::Error(
             _logger,
-            "Chonology error: Received NextSimTask from participant \'{}\' with lower timePoint {} than last "
+            "Chronology error: Received NextSimTask from participant \'{}\' with lower timePoint {} than last "
```
VDanielEdwards:
```suggestion
            "Chronology error: Received NextSimTask from participant '{}' with lower timePoint {} than last "
```

SilKit/source/core/participant/Participant_impl.hpp
```diff
@@ -1328,7 +1328,7 @@ template <typename SilKitConnectionT>
 template <typename MessageT>
 void Participant<SilKitConnectionT>::HandleSynchronizationPoint()
 {
-    if constexpr (SilKitMsgTraits<typename RemoveCvRef<MessageT>::type>::IsSynchronizationPoint())
+    if constexpr (SilKitMsgTraits<typename RemoveCvRef<MessageT>>::IsSynchronizationPoint())
```
VDanielEdwards: You should also not need the `typename`, since the compiler knows that `RemoveCvRef` is a type.

```suggestion
    if constexpr (SilKitMsgTraits<RemoveCvRef<MessageT>>::IsSynchronizationPoint())
```

MariusBgm:
closing this. we will work on the concept in a different PR

PR #275, opened 2025-11-14
SilKitMonitor: add appropriate default log level

VJanKraemer:
Since we changed the default log level, the system monitor would not output any info if not given a dedicated config. This changes the default config from an empty config to a config specifying "Info" as the log level.

Issues: SILKIT-1884

### DIFF
--- a/Utilities/SilKitMonitor/PassiveSystemMonitor.cpp
+++ b/Utilities/SilKitMonitor/PassiveSystemMonitor.cpp
@@ -35,6 +35,26 @@ std::ostream& operator<<(std::ostream& out, std::chrono::nanoseconds timestamp)
     return out;
 }
 
+auto getConfig(const std::string& configurationFilename, const std::string& logLevel)
+{
+    if (!configurationFilename.empty())
+    {
+        if (!logLevel.empty())
+        {
+            std::cout << "warning: loglevel provided via commandline will be ignored in favor of the provided "
+                         "participant config!"
+                      << std::endl;
+        }
+        return SilKit::Config::ParticipantConfigurationFromFile(configurationFilename);
+    }
+    else
+    {
+        std::stringstream defaultConfig;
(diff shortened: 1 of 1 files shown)


Utilities/SilKitMonitor/PassiveSystemMonitor.cpp
```diff
@@ -141,6 +167,13 @@ int main(int argc, char** argv)
         auto* logger = participant->GetLogger();
         auto* systemMonitor = participant->CreateSystemMonitor();
 
+        systemMonitor->SetParticipantConnectedHandler(
+            [logger](const Services::Orchestration::ParticipantConnectionInformation& status) {
+            std::stringstream buffer;
+            buffer << "New Participan: \'" << status.participantName << std::endl;
```
KonradBreitsprecherBkd: Typo + missing tick
```suggestion
            buffer << "New Participant: \'" << status.participantName << "\'" << std::endl;
```

Utilities/SilKitMonitor/PassiveSystemMonitor.cpp
```diff
@@ -47,6 +67,11 @@ int main(int argc, char** argv)
         "name", "n", "SystemMonitor", "[--name <participantName>]",
         "-n, --name <participantName>: The participant name used to take part in the simulation. Defaults to "
         "'SystemMonitor'.");
+    commandlineParser.Add<CliParser::Option>("loglevel", "l", "",
+                                             "[--loglevel <Critical|Error|Warning|Info|Debug|Trace>]",
+                                             "-l, --loglevel <Critical|Error|Warning|Info|Debug|Trace>: The log level "
+                                             "used for the sil-kit-monitor. Defaults to "
+                                             "INFO.");
```
KonradBreitsprecherBkd: This is case sensitive, so it should be "Info".
```suggestion
                                             "Info.");
```

PR #283, opened 2025-12-11
cmake: Enable Dashboard in Linux package builds

VJanKraemer:
Add the Dashboard back to all Linux Packages (deb/rpm)


MariusBgm (approved):
LGTM!

PR #284, opened 2025-12-19
Add SetStepDuration and TimeAdvanceModes

KonradBreitsprecherBkd:
This PR adds support for dynamic time step management in the SIL Kit orchestration layer by introducing `SetStepDuration `and `TimeAdvanceMode `features. The changes enable participants to either advance time based on their own step duration or synchronize with the time points among all participants.

**Key Changes:**
- Introduced a new `TimeAdvanceMode` enum with two modes: `ByOwnDuration` and `ByMinimalDuration`
- Renamed the internal `SetPeriod` method to `SetStepDuration` and exposed it in the public API
- Added a new overload of `CreateTimeSyncService` that accepts a `TimeAdvanceMode` parameter

With these changes, a participant can reduce the step duration in a critical phase where a more fine grained synchronization is needed.

### DIFF
--- a/SilKit/source/services/orchestration/TimeConfiguration.cpp
+++ b/SilKit/source/services/orchestration/TimeConfiguration.cpp
@@ -101,52 +101,72 @@ void TimeConfiguration::OnReceiveNextSimStep(const std::string& participantName,
 void TimeConfiguration::SetStepDuration(std::chrono::nanoseconds duration)
 {
     Lock lock{_mx};
+
+    if (duration == 0ns)
+    {
+        throw SilKitError("Attempted to set step duration to zero.");
+    }
+
     _myNextTask.duration = duration;
 }
 
-auto TimeConfiguration::GetMinimalOtherDuration() const -> std::chrono::nanoseconds
+auto TimeConfiguration::GetMinimalAlignedDuration() const -> std::chrono::nanoseconds
 {
-    Lock lock{_mx};
     if (_otherNextTasks.empty())
     {
-        return std::chrono::nanoseconds{0};
+        return std::chrono::nanoseconds::max();
     }
 
-    auto minDuration = std::chrono::nanoseconds::max();
+    auto earliestOtherTimepoint = std::chrono::nanoseconds::max();
     for (const auto& entry : _otherNextTasks)
     {
-        if (entry.second.duration < minDuration)
+        // Both start and end of other participant's step could be the earliest next timepoint
+        auto nextStepStart = entry.second.timePoint;
+        auto nextStepEnd = entry.second.timePoint + entry.second.duration;
+        if (nextStepStart > _currentTask.timePoint && nextStepStart < earliestOtherTimepoint)
         {
-            minDuration = entry.second.duration;
+            earliestOtherTimepoint = nextStepStart;
+        }
+        else if (nextStepEnd < earliestOtherTimepoint)
+        {
+            earliestOtherTimepoint = nextStepEnd;
         }
     }
-    return minDuration;
-}
+    //Logging::Info(_logger, "Earliest next timepoint among other participants is {}ms",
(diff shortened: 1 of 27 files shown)


SilKit/source/services/orchestration/TimeConfiguration.cpp
```diff
@@ -82,36 +92,61 @@ void TimeConfiguration::OnReceiveNextSimStep(const std::string& participantName,
...
+    if (_otherNextTasks.empty())
     {
-        const std::string errorMessage{"Participant " + otherParticipantName + " unknown."};
-        throw SilKitError{errorMessage};
+        return std::chrono::nanoseconds{0};
     }
-    auto it = _otherNextTasks.find(otherParticipantName);
-    if (it != _otherNextTasks.end())
+
+    auto minDuration = std::chrono::nanoseconds::max();
```
VDanielEdwards: Are you sure using the minimum of the other sim. step durations works correctly if not all sim. step sizes are multiples of the smalles sim. step size? Especially if the 'minimal-duration' participant performs a hop-on and lands on an 'odd' current time.

Example with two participants `A` and `B`, the numbers are the 'now' timestamps, ordered as they would occur in SIL Kit (same x position means these steps don't wait for each other):

```
A: 0 2   4 6 8   10 12 14    16 18
B: 0   3   6   9    12    15    18
```

A participant `C` that uses the mimimum-duration mode added to the next run of that simulation, would do the same as participant `A`. Is this the desired behavior?


```
A: 0 2   4 6 8   10 12 14    16 18
B: 0   3   6   9    12    15    18
C: 0 2   4 6 8   10 12 14    16 18
```

To the best of my knowledge, if the participant `C` uses hop-on and joins a running simulation the following could happen (if `C` joins such that it's first `now` is taken from `B`):

```
A: 0 2   4 6 8   10    12    14    16    18
B: 0   3   6   9       12       15       18
C:             9    11    13    15    17
```

My initial feeling is that the selected duration should be computed as the difference between the 'current time' (of the minimal-duration participant) and the earliest 'next time' across all other 'normal' participants?

This would lead to the following in the starting-together case:

```
A: 0 2   4 6 8   10 12 14    16 18
B: 0   3   6   9    12    15    18
C: 0 2 3 4 6 8 9 10 12 14 15 16 18
```

And the hop-on case:

```
A: 0 2   4 6 8   10 12 14    16 18
B: 0   3   6   9    12    15    18
C:             9 10 12 14 15 16 18
```

Notice that they behave the exact same way in this case.

KonradBreitsprecherBkd: Thanks @VDanielEdwards, you're right. Only adding the min. duration can lead to unaligned 'grids'. I will change the calculation such that all participants configured with `ByMinimalDuration` are guaranteed to trigger the SimStep with the same time.

SilKit/source/services/orchestration/TimeConfiguration.cpp
```diff
@@ -82,35 +92,80 @@ void TimeConfiguration::OnReceiveNextSimStep(const std::string& participantName,
...
+    {
+        return std::chrono::nanoseconds::max();
+    }
+
+    auto earliestOtherTimepoint = std::chrono::nanoseconds::max();
+    for (const auto& entry : _otherNextTasks)
+    {
+        // Both start and end of other participant's step could be the earliest next timepoint
+        auto nextStepStart = entry.second.timePoint;
+        auto nextStepEnd = entry.second.timePoint + entry.second.duration;
```
VDanielEdwards:
```suggestion
        const auto nextStepStart = entry.second.timePoint;
        const auto nextStepEnd = entry.second.timePoint + entry.second.duration;
```

These two locals can be `const` as they aren't modified.

KonradBreitsprecherBkd:
Closing this in favor of https://github.com/vectorgrp/sil-kit/pull/352

Issue #286, opened 2026-01-08
Unclear handling of CAN message size in virtual CAN networks


JannisRln:
**Title**
Unclear handling of CAN message size in virtual CAN networks

**Description**
While integrating virtual CAN networks using SIL Kit, it is unclear how CAN message size constraints are handled by the framework. The current API documentation does not clearly specify whether message size validation, truncation, or error handling is performed internally, or whether this responsibility is delegated entirely to the user.

This ambiguity makes it difficult to design robust applications that rely on virtual CAN communication and to ensure consistency with real CAN/CAN FD behavior.

**Questions / Points needing clarification**

1. What is the maximum supported payload size for CAN and CAN FD messages in virtual CAN networks?
2. Does SIL Kit validate message size against CAN/CAN FD limits when transmitting?
3. If an oversized message is sent: - Is it rejected with an error? - Silently truncated? - Passed through unchanged?
4. Where in the API or documentation is this behavior formally defined?

**Expected outcome**
- Clear documentation describing how message size is handled for virtual CAN networks.
- Explicit guarantees about validation, error handling, and consistency with CAN/CAN FD specifications.
- (Optional) Example code or API references demonstrating correct usage of large messages.

**Environment**

SIL Kit version: 5.0.3
OS: Windows 64bit 
Language bindings: C API and C++ API used in python with pybind11 (https://github.com/JannisRln/SilKit_py)

PR #288, opened 2026-01-09
Document CAN message size handling

KonradBreitsprecherBkd:
## Subject
Improves documentation related to https://github.com/vectorgrp/sil-kit/issues/286.

### DIFF
--- a/docs/api/services/can.rst
+++ b/docs/api/services/can.rst
@@ -196,7 +196,7 @@ Enumerations and Typedefs
 Usage Examples
 --------------
 
-This section contains complete examples that demonstrate the usage of the CAN controller and the interaction between
+This section contains complete examples that demonstrate the use of the CAN controller and the interaction between
 two or more controllers.
 
 Although the CAN controllers would typically belong to different participants and reside in different 


docs/api/services/can.rst
```diff
@@ -189,9 +196,11 @@ Enumerations and Typedefs
 Usage Examples
 --------------
 
-This section contains complete examples that show the usage of the CAN controller and the interaction of two or more 
-controllers. Although the CAN controllers would typically belong to different participants and reside in different
-processes, their interaction is shown sequentially to demonstrate cause and effect.
+This section contains complete examples that demonstrate the usage of the CAN controller and the interaction between
```
VJanKraemer: I checked and I think that use is actual "more" correct version. Usage should be used for either "general accepted/habitual use of something" or if mode or amounts of using is relevant https://www.merriam-webster.com/dictionary/usage

Nitpicking 🤣  but maybe the more general use would be better applicable.

VJanKraemer (approved):
Change use vs usage at your own leisure since we established that both forms are valid.
LGTM :)

PR #291, opened 2026-01-21
Set Demo and Utility `RPATH`s to match system defaults

snps-behrens:
## Description

When a SIL Kit installation is built, the system default is used to name the lib-directory in the install hierarchy. On certain Linux distributions (Alma Linux, for instance) this default lib-directory is `lib64` instead of `lib`. This results in the SIL Kit Demos and Utilities not being able to find `libSilKit.so` because their RPATH is hard-coded to use `lib`. 

This PR proposes setting these `RPATH`s to match the naming scheme used when creating the SIL Kit installation (as defined by the `GNUInstallDirs` CMake module).

- [x] Changelog.md updated
- [X] Documentation updated
- [x] Rebase &rarr; commit history clean

### DIFF
--- a/docs/changelog/versions/latest.md
+++ b/docs/changelog/versions/latest.md
@@ -3,3 +3,5 @@
 ## Fixed
 
 - `sil-kit-registry`: fixed the configurability of the listen URI, when used as Windows System Service.
+
+- `cmake`: Demo and Utility `RPATH`s are now set to match the install hierarchy on all systems.


Utilities/SilKitMonitor/CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ add_executable(sil-kit-monitor
 set_target_properties(sil-kit-monitor PROPERTIES
      FOLDER "Utilities"
      RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/$<CONFIG>"
-     INSTALL_RPATH "$ORIGIN/../lib"
+     INSTALL_RPATH "$ORIGIN/../${CMAKE_INSTALL_LIBDIR}"
```
snps-behrens, line 21: Just a question that occurred to me looking at this again: Is there a reason the `sil-kit-monitor` does not have the trailing `$ORIGIN` that the other executables have? Or maybe the other way around; what is the reason that all the other executables are installed with a trailing `$ORIGIN` in the `RPATH`?

VJanKraemer: This was before my time, so I don't know the exact reason, but the whole RPATH shenanigans are happening so we/users can take the SIL Kit Zip distributions and just copy them wherever on the target system and as long as the folder hierarchy of the zip is kept intact, everything works. The trailing $ORIGIN might be a remnant from earlier distribution models, but afaik it is not needed in the current distribution model. Can't be sure though to change that without breaking some users workflow. 

To be frank I'd rather get rid of the whole rpath setting by default via CMakeLists.txt and put it specifically into the CMake cmdline/scripts that build the ZIP distributions. But that is a bigger discussion to be had, so for the moment I am just glad someone fixed it for non Debian derived Linux distributions :)

PR #297, opened 2026-02-17
Populate `timestamp` in `LinControllerStatusUpdate` message

snps-kratz:
## Subject

Initializes the `timestamp` variable on the `LinControllerStatusUpdate` message send in `LinController::SetControllerStatusInternal`.
Adds test conditions that check the timestamp of the `LinControllerStatusUpdate` message.

## Description

This change fixes inconsistent logging behavior for trace logs. Originally, the timestamp field was uninitialized, leading to trace logs with timestamps that would not match simulation timestamps.
Additionally, this change populates the timestamp received by network simulators in `OnControllerStatusUpdate` function calls.

Since the tests have been changed, correctness in those should be focused.

- [x] Prepared update for depending repositories
- [x] API docstrings updated (public API changes only)
- [x] Rebase &rarr; commit history clean

### DIFF
--- a/SilKit/source/services/lin/LinController.cpp
+++ b/SilKit/source/services/lin/LinController.cpp
@@ -606,7 +606,7 @@ void LinController::SetControllerStatusInternal(LinControllerStatus status)
 
     _controllerStatus = status;
 
-    LinControllerStatusUpdate msg;
+    LinControllerStatusUpdate msg{};
     msg.status = status;
     msg.timestamp = _timeProvider->Now();
 


SilKit/source/services/lin/LinController.cpp
```diff
@@ -608,6 +608,7 @@ void LinController::SetControllerStatusInternal(LinControllerStatus status)
 
     LinControllerStatusUpdate msg;
```
VDanielEdwards:
```suggestion
    LinControllerStatusUpdate msg{};
```

PR #313, opened 2026-04-24
style: use full paths for internal headers in `capi/...`

VDanielEdwards:
## Subject

Use full paths for internal headers in all sources and headers in `SilKit/source/capi/`. Rearranges the includes to follow the guidelines in the [LLVM Styleguide for C++](https://llvm.org/docs/CodingStandards.html#include-style) ~~[Google Styleguide for C++](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes)~~. Removes unneccessary includes.

- [ ] Changelog.md updated

### DIFF
--- a/SilKit/source/capi/CapiOrchestration.cpp
+++ b/SilKit/source/capi/CapiOrchestration.cpp
@@ -2,22 +2,21 @@
 //
 // SPDX-License-Identifier: MIT
 
-#include "silkit/capi/SilKit.h"
-#include "silkit/SilKit.hpp"
-#include "silkit/services/orchestration/all.hpp"
-#include "silkit/participant/exception.hpp"
+#include "silkit/capi/Orchestration.h"
 
+#include "capi/CapiImpl.hpp"
 #include "participant/ParticipantExtensionsImpl.hpp"
-
-#include "CapiImpl.hpp"
-#include "TypeConversion.hpp"
-
 #include "services/orchestration/TimeSyncServiceExtensionsImpl.hpp"
 
-#include <memory>
+#include "silkit/experimental/services/orchestration/ISystemController.hpp"
+#include "silkit/participant/IParticipant.hpp"
+#include "silkit/services/orchestration/all.hpp"
+
+#include <chrono>
+#include <future>
 #include <map>
-#include <mutex>
-#include <cstring>
+#include <memory>
+#include <string>
 
 
 namespace {
--- a/SilKit/source/capi/CapiParticipant.cpp
+++ b/SilKit/source/capi/CapiParticipant.cpp
@@ -2,23 +2,18 @@
 //
 // SPDX-License-Identifier: MIT
 
-#include "ParticipantConfiguration.hpp"
-#include "ParticipantConfigurationFromXImpl.hpp"
+#include "silkit/capi/Participant.h"
+
 #include "CreateParticipantImpl.hpp"
-#include "YamlParser.hpp"
+#include "capi/CapiImpl.hpp"
+#include "config/ParticipantConfiguration.hpp"
+#include "config/ParticipantConfigurationFromXImpl.hpp"
+#include "config/YamlParser.hpp"
 
-#include "silkit/capi/SilKit.h"
-#include "silkit/SilKit.hpp"
-#include "silkit/services/logging/ILogger.hpp"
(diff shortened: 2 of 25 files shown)


snps-behrens:
Is there any particular reason you chose the google style guide? I have found that using the [llvm style guide](https://llvm.org/docs/CodingStandards.html#include-style) leads to headers being more self-contained.

VDanielEdwards:
> Is there any particular reason you chose the google style guide? I have found that using the [llvm style guide](https://llvm.org/docs/CodingStandards.html#include-style) leads to headers being more self-contained.

No particular reason, it was just the first that came to mind. But I do like the ordering from local -> global as in the LLVM styleguide more than what the Google styleguide suggests.

@MariusBgm would you object to switching the include-order to the LLVM styleguide?

SilKit/source/capi/CapiOrchestration.cpp
```diff
@@ -2,22 +2,17 @@
...
-#include "TypeConversion.hpp"
-
 #include "services/orchestration/TimeSyncServiceExtensionsImpl.hpp"
 
-#include <memory>
+#include "silkit/experimental/services/orchestration/ISystemController.hpp"
+#include "silkit/participant/IParticipant.hpp"
+#include "silkit/services/orchestration/all.hpp"
+
 #include <map>
```
snps-behrens, line 17: Why remove mutex here? It's still used in this file.

SilKit/source/capi/CapiParticipant.cpp
```diff
@@ -2,23 +2,15 @@
...
+#include "capi/CapiImpl.hpp"
+#include "config/ParticipantConfiguration.hpp"
+#include "config/ParticipantConfigurationFromXImpl.hpp"
+#include "config/YamlParser.hpp"
 
-#include <memory>
-#include <map>
-#include <mutex>
-#include <fstream>
+#include "silkit/participant/IParticipant.hpp"
```
snps-behrens, line 14: Why remove `memory` here?

SilKit/source/capi/CapiVendor.cpp
```diff
@@ -2,20 +2,14 @@
...
-#include "ParticipantConfiguration.hpp"
-#include "CreateSilKitRegistryImpl.hpp"
+#include "silkit/capi/Vendor.h"
 
-#include "silkit/capi/SilKit.h"
+#include "CreateSilKitRegistryImpl.hpp"
+#include "capi/CapiImpl.hpp"
+#include "config/ParticipantConfiguration.hpp"
 
 #include "silkit/vendor/ISilKitRegistry.hpp"
```
snps-behrens, line 11: same here (`memory`)?

SilKit/source/capi/Test_CapiRpc.cpp
```diff
@@ -6,12 +6,15 @@
 #define _CRT_SECURE_NO_WARNINGS
```
snps-behrens, line 6: Not really in scope for this PR, but would it make sense to set this flag globally, rather than in every file where these warnings trigger (or have triggered in the past)?

e.g. by adding something like `$<$<COMPILE_LANG_AND_ID:CXX,MSVC>:_CRT_SECURE_NO_WARNINGS>` to the global `add_compile_definitions`.

snps-behrens (commented):
Looks pretty good. I did notice you removed pretty much all system includes from these files, even where there is extensive use of the standard library. Is that intentional?

SilKit/source/capi/Test_CapiRpc.cpp
VDanielEdwards: I created an issue in the planning project for that. Thank you!

VDanielEdwards:
> Looks pretty good. I did notice you removed pretty much all system includes from these files, even where there is extensive use of the standard library. Is that intentional?

I got a bit overzealous there I think. Re-added in the latest fixup commit. Good catch!

VDanielEdwards:
The build error is resolved on main. If you don't mind, I would rebase this branch @snps-behrens ?

snps-behrens:
Go ahead and rebase. Thanks for warning me. 👍

VDanielEdwards:
Closed in favor of Marius's changes.

PR #325, opened 2026-05-08
demo: remove third-party dependencies from performance test script

VDanielEdwards:
## Subject

Two more Dependabot alerts regarding GitPython caused a PR in the internal mirror. This should fix those.

image

- [ ] Changelog.md updated

### DIFF
--- a/Demos/tools/Benchmark/requirements.txt
+++ b/Demos/tools/Benchmark/requirements.txt
@@ -1,3 +1,1 @@
-gitdb==4.0.12
-GitPython==3.1.47
-smmap==5.0.2
+# no third-party requirements
--- a/Demos/tools/Benchmark/PerformanceTests.py
+++ b/Demos/tools/Benchmark/PerformanceTests.py
@@ -18,8 +18,6 @@ import csv
 import argparse
 import typing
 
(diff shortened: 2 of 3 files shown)


Demos/tools/Benchmark/requirements.txt
```diff
@@ -1,3 +1,3 @@
 gitdb==4.0.12
-GitPython==3.1.47
+GitPython==3.1.49
```
VJanKraemer: It semms it has been updated to 3.1.50 now. Maybe medium term we should get rid of it entirely?
https://github.com/vectorgrp/sil-kit/security/dependabot/3


```suggestion
GitPython==3.1.50
```

VJanKraemer (changes requested):

Demos/tools/Benchmark/requirements.txt
VDanielEdwards: Done, replaced with `subprocess.run` expecting that `git` is on the path. Also fixes some bugs and allows extra arguments to the cmake configuration step, e.g., to select `Ninja` as a generator.

VJanKraemer (approved):
LGTM

PR #339, opened 2026-06-11
build: commit SilKitVersionMacros.h and regenerate with tool

MariusBgm:
Commit the version macros directly to the public include folders.
Add a small C++17 tool to regenerate the header with new version, git hash and optionally rotate the changelog.md if version number changed.
the tool does:
- update `SilKit/cmake/SilKitVersion.cmake`
- update `SilKit/include/silkit/capi/SilKitVersionMacros.h` with git hash of last commit
- rotate the changelog if the version changes

also delete JenkinsFile and do not use SILKIT_BUILD_NUMBER in the versioning scheme


MariusBgm:
i think we should be able to pass the git hash as a define to the compilation

MariusBgm:
> i think we should be able to pass the git hash as a define to the compilation

i think something like:
```
# CI: update the header in-tree (not committed, just for this build)
./sil-kit-version-gen \
  --major 5 --minor 0 --patch 6 \
  --git-hash "$(git rev-parse HEAD)" \
  --no-changelog \
  SilKit/include/silkit/capi/SilKitVersionMacros.h
```
would work in CI builds, without changing the current implementation.
and in bazel we could probably create it using something like:
```
# tools/BUILD.bazel — build the generator itself
cc_binary(
    name = "sil-kit-version-gen",
    srcs = ["//SilKit/source/util:sil-kit-version-gen.cpp"],
    # C++17, no deps — trivially hermetic
)

# SilKit/BUILD.bazel — generate the header
genrule(
    name = "version_macros",
    srcs = [".git/HEAD"],   # or nothing if hash is passed in
    outs = ["include/silkit/capi/SilKitVersionMacros.h"],
    cmd = "$(location //tools:sil-kit-version-gen) " +
          "--major 5 --minor 0 --patch 6 " +
          "--git-hash $$(cat bazel-out/stable-status.txt | grep GIT_HASH | cut -d' ' -f2) " +
          "--no-changelog $@",
    tools = ["//tools:sil-kit-version-gen"],
    stamp = 1,  # gives the rule access to stable-status.txt
)

cc_library(
    name = "version_header",
    hdrs = [":version_macros"],
    includes = ["include"],
)
```

PR #341, opened 2026-06-12
Add Integration Test for Timestamp Behavior

KonradBreitsprecherBkd:
The [timestamp behavior](https://vectorgrp.github.io/sil-kit-docs/simulation/simulation.html#timestamps-in-messages) had no integration test. The test covers:

Verify timestamp behavior from the synchronized participant perspective:
- Sync sender: outgoing event carries virtual send time.
- Async sender: outgoing events will have an invalid timestamp
- Sync receiver: incoming event from async or sync sender is timestamped with sync receive time.
- Async receiver: incoming event from async or sync sender keeps the timestamp (async sender: invalid timestamp, sync sender: send time)

### DIFF
--- a/SilKit/IntegrationTests/ITest_AsyncSimTask.cpp
+++ b/SilKit/IntegrationTests/ITest_AsyncSimTask.cpp
@@ -420,8 +420,8 @@ TEST(ITest_AsyncSimTask, test_timestamp_handling)
                                                    const SilKit::Services::PubSub::DataMessageEvent& dataMessageEvent) {
         const auto currentStep = std::chrono::nanoseconds{syncCurrentStepNs.load()};
 
-        EXPECT_TRUE(dataMessageEvent.timestamp == currentStep)
-            << "Incoming async message should be timestamped with the sync receiver time.";
+        EXPECT_TRUE((dataMessageEvent.timestamp == currentStep) || (dataMessageEvent.timestamp == currentStep + 1ms))
+            << "Incoming async message should be timestamped with the sync receiver time (or +1ms tolerance).";
         asyncToSyncReceived = true;
     });
 
@@ -431,8 +431,8 @@ TEST(ITest_AsyncSimTask, test_timestamp_handling)
                                                   const SilKit::Services::PubSub::DataMessageEvent& dataMessageEvent) {
         const auto currentStep = std::chrono::nanoseconds{syncCurrentStepNs.load()};
 
-        EXPECT_TRUE(dataMessageEvent.timestamp == currentStep)
-            << "Incoming sync message should be timestamped with the sync receiver time.";
+        EXPECT_TRUE((dataMessageEvent.timestamp == currentStep) || (dataMessageEvent.timestamp == currentStep + 1ms))
+            << "Incoming sync message should be timestamped with the sync receiver time (or +1ms tolerance).";
         syncToSyncReceived = true;
     });
 
--- a/SilKit/IntegrationTests/ITest_AsyncSimTask.cpp
+++ b/SilKit/IntegrationTests/ITest_AsyncSimTask.cpp
@@ -365,11 +365,17 @@ TEST(ITest_AsyncSimTask, test_async_simtask_other_simulation_steps_completed_han
 // Verify timestamp behavior from the synchronized participant perspective:
 // - Sync sender: outgoing event carries virtual send time.
 // - Async sender: outgoing events will have an invalid timestamp
-// - Sync receiver: incoming event from async or sync sender is timestamped with sync receive time.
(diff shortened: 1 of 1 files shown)


snps-behrens (commented):
Just to clarify. I understood the timestamp behavior implementation in the sync -> sync situation to be that the send time (not the receive time) is used. If it were otherwise, wouldn't NetworkSimulator timestamps be overwritten?

The documentation seems to agree with me (though it is not explicilty stated): https://vectorgrp.github.io/sil-kit-docs/simulation/simulation.html#timestamps-in-messages.

KonradBreitsprecherBkd:
> Just to clarify. I understood the timestamp behavior implementation in the sync -> sync situation to be that the send time (not the receive time) is used. If it were otherwise, wouldn't NetworkSimulator timestamps be overwritten?
> 
> The documentation seems to agree with me (though it is not explicilty stated): https://vectorgrp.github.io/sil-kit-docs/simulation/simulation.html#timestamps-in-messages.

You're right, the logic is (taken from the code):
```
template <typename MsgT>
void SetTimestamp(MsgT& msg, std::chrono::nanoseconds value, std::enable_if_t<HasTimestamp<MsgT>::value, bool> = true)
{
    if (msg.timestamp == std::chrono::nanoseconds::duration::min())
    {
        msg.timestamp = value;
    }
}
```
This is: "On reception of a msg with an invalid timestamp (==from async), overwrite it with the local now() of the receiver"

So the comment above the test is misleading:
` - Sync receiver: incoming event from async or sync sender is timestamped with sync receive time.`
Also the `SyncSubFromSync` evaluation is off then.
Good catch, will rework!

SilKit/IntegrationTests/ITest_AsyncSimTask.cpp
```diff
@@ -362,4 +362,140 @@ TEST(ITest_AsyncSimTask, test_async_simtask_other_simulation_steps_completed_han
...
+    auto* async2Participant = testHarness.GetParticipant("Async2");
+
+    auto* syncLifecycleService = syncParticipant->GetOrCreateLifecycleService();
+    auto* sync2LifecycleService = sync2Participant->GetOrCreateLifecycleService();
+    auto* asyncLifecycleService = asyncParticipant->GetOrCreateLifecycleService();
+    auto* async2LifecycleService = async2Participant->GetOrCreateLifecycleService();
+    auto* syncTimeSyncService = syncParticipant->GetOrCreateTimeSyncService();
+    auto* sync2TimeSyncService = sync2Participant->GetOrCreateTimeSyncService();
+
+    auto* syncPublisher = syncParticipant->Participant()->CreateDataPublisher("SyncPub", dataSpecSyncToAsync);
```
snps-behrens: nitpick: would it make sense to rename this and the next one to `syncPublisherToAsync` and `asyncPublisherToSync` to make the test a little more self-documenting?

SilKit/IntegrationTests/ITest_AsyncSimTask.cpp
```diff
@@ -362,4 +362,140 @@ TEST(ITest_AsyncSimTask, test_async_simtask_other_simulation_steps_completed_han
...
+
+    syncParticipant->Participant()->CreateDataSubscriber(
+        "SyncSub", dataSpecAsyncToSync,
+        [&syncCurrentStepNs, &asyncToSyncReceived](SilKit::Services::PubSub::IDataSubscriber* /*subscriber*/,
+                                                   const SilKit::Services::PubSub::DataMessageEvent& dataMessageEvent) {
+        const auto currentStep = std::chrono::nanoseconds{syncCurrentStepNs.load()};
+
+        EXPECT_TRUE((dataMessageEvent.timestamp == currentStep) || (dataMessageEvent.timestamp == currentStep + 1ms)
+                    || (dataMessageEvent.timestamp == currentStep - 1ms))
+            << "Incoming async message should be timestamped with the sync receiver time (+/-1ms tolerance).";
```
snps-behrens: Maybe I'm missing something, but isn't this handler run as part of the same communication thread job that sets the timestamp? How is it possible for `dataMessageEvent.timestamp != currentStep`?

snps-behrens (commented):
Looks good to me!

I would like to clear up one question I had (might just be down to me not having looked into the SilKit time synchronization internals in a long time).

SilKit/IntegrationTests/ITest_AsyncSimTask.cpp
KonradBreitsprecherBkd: You're right, the tolerance is not needed here.

snps-behrens (approved):

PR #344, opened 2026-06-15
Add Integration Test for Remote Logging

KonradBreitsprecherBkd:
## Summary
Add a new integration test for remote logging with three participants in `SilKit/IntegrationTests/ITest_RemoteLogging.cpp`.

- Creates three participants:
  - `Sender1` and `Sender2` configured with a remote logging sink
  - `Receiver` configured with `LogFromRemotes: true` and a file sink
- `Sender1` and `Sender2` each emit a unique log message
- Verifies that the receiver-side log file contains both messages
- Timeout-based polling for file creation and content visibility
- Added scoped cleanup to remove generated log files after test execution (including failure paths)

### DIFF
--- a/SilKit/IntegrationTests/ITest_RemoteLogging.cpp
+++ b/SilKit/IntegrationTests/ITest_RemoteLogging.cpp
@@ -6,7 +6,6 @@
 #include <chrono>
 #include <filesystem>
 #include <fstream>
-#include <sstream>
 #include <string>
 #include <thread>
 #include <vector>
@@ -16,6 +15,7 @@
 #include "SimTestHarness.hpp"
 
 #include "gtest/gtest.h"
+#include "gmock/gmock.h"
 
 namespace {
 
@@ -47,18 +47,6 @@ auto FindLogFiles(const std::string& logNamePrefix) -> std::vector<std::filesyst
     }
 
     return candidates;
-}
-
-auto FindLogFile(const std::string& logNamePrefix) -> std::filesystem::path
-{
-    const auto candidates = FindLogFiles(logNamePrefix);
-
-    if (candidates.size() != 1u)
-    {
-        return {};
-    }
-
-    return candidates.front();
 }
 
 struct ScopedLogFileCleanup
@@ -163,29 +151,12 @@ Logging:
 
     ASSERT_TRUE(testHarness.Run(5s));
 
-    const auto deadline = std::chrono::steady_clock::now() + 10s;
-    std::filesystem::path logFile;
-    std::string logContent;
+    const auto logFiles = FindLogFiles(filePrefix);
+    ASSERT_EQ(logFiles.size(), 1u) << "Expected exactly one receiver log file with prefix " << filePrefix;
 
-    while (std::chrono::steady_clock::now() < deadline)
-    {
-        logFile = FindLogFile(filePrefix);
-        if (!logFile.empty())
-        {
-            logContent = ReadTextFile(logFile);
-            if (logContent.find(sender1Message) != std::string::npos
-                && logContent.find(sender2Message) != std::string::npos)
-            {
-                break;
-            }
-        }
-
-        std::this_thread::sleep_for(10ms);
-    }
-
-    ASSERT_FALSE(logFile.empty()) << "Could not find exactly one receiver log file with prefix " << filePrefix << "\n";
-    EXPECT_NE(logContent.find(sender1Message), std::string::npos);
-    EXPECT_NE(logContent.find(sender2Message), std::string::npos);
+    const auto logContent = ReadTextFile(logFiles.front());
+    EXPECT_THAT(logContent, testing::HasSubstr(sender1Message));
+    EXPECT_THAT(logContent, testing::HasSubstr(sender2Message));
 }
 
 } // namespace
--- a/SilKit/IntegrationTests/ITest_RemoteLogging.cpp
+++ b/SilKit/IntegrationTests/ITest_RemoteLogging.cpp
@@ -1,9 +1,8 @@
-// SPDX-FileCopyrightText: 2022 Vector Informatik GmbH
+// SPDX-FileCopyrightText: 2026 Vector Informatik GmbH
 //
 // SPDX-License-Identifier: MIT
 
-#include <algorithm>
-#include <atomic>
+#include <system_error>
 #include <chrono>
 #include <filesystem>
 #include <fstream>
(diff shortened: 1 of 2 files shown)


SilKit/IntegrationTests/ITest_RemoteLogging.cpp
```diff
@@ -0,0 +1,178 @@
+// SPDX-FileCopyrightText: 2022 Vector Informatik GmbH
+//
+// SPDX-License-Identifier: MIT
+
+#include <algorithm>
```
snps-behrens: `algorithm` is unused, and `system_error` is not included.

SilKit/IntegrationTests/ITest_RemoteLogging.cpp
```diff
@@ -0,0 +1,178 @@
+// SPDX-FileCopyrightText: 2022 Vector Informatik GmbH
```
snps-behrens: Time flies, I think it's 2026 already. 😉

SilKit/IntegrationTests/ITest_RemoteLogging.cpp
```diff
@@ -0,0 +1,178 @@
...
+    const auto receiverConfig = "Logging:\n"
+                                "  LogFromRemotes: true\n"
+                                "  FlushLevel: Trace\n"
+                                "  Sinks:\n"
+                                "    - Type: File\n"
+                                "      Level: Trace\n"
+                                "      LogName: "
+        + receiverLogName + "\n";
+
+    SimTestHarness testHarness{{"Sender1", "Sender2", "Receiver"}, "silkit://localhost:0", true};
```
snps-behrens: This version of the constructor is marked deprecated, and says the single-argument version taking a `SimTestHarnessArgs` object should be used.

SilKit/IntegrationTests/ITest_RemoteLogging.cpp
```diff
@@ -0,0 +1,178 @@
...
+    ScopedLogFileCleanup cleanup{filePrefix};
+
+    const auto senderConfig = R"(
+Logging:
+  Sinks:
+    - Type: Remote
+      Level: Trace
+)";
+
+    const auto receiverConfig = "Logging:\n"
```
snps-behrens: I think this could be made more consistent with the above `senderConfig` by using a raw string inside a call to `fmt::format()`.

SilKit/IntegrationTests/ITest_RemoteLogging.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+    receiverTimeSync->SetSimulationStepHandler([](std::chrono::nanoseconds /*now*/, std::chrono::nanoseconds) {}, 1ms);
+
+    ASSERT_TRUE(testHarness.Run(5s));
+
+    const auto deadline = std::chrono::steady_clock::now() + 2s;
+    std::filesystem::path logFile;
+    std::string logContent;
+
+    while (std::chrono::steady_clock::now() < deadline)
```
snps-behrens: Why is this polling loop necessary?

SilKit/IntegrationTests/ITest_RemoteLogging.cpp
```diff
@@ -0,0 +1,178 @@
...
+    auto* sender2TimeSync = sender2->GetOrCreateTimeSyncService();
+    auto* receiverTimeSync = receiver->GetOrCreateTimeSyncService();
+
+    auto* sender1Logger = sender1->Participant()->GetLogger();
+    auto* sender2Logger = sender2->Participant()->GetLogger();
+
+    const std::string sender1Message = "remote-log-from-sender-1";
+    const std::string sender2Message = "remote-log-from-sender-2";
+
+    std::atomic<bool> sender1Logged{false};
```
snps-behrens: I don't think these `std::atomic`s are needed, right?

SilKit/IntegrationTests/ITest_RemoteLogging.cpp
```diff
@@ -0,0 +1,178 @@
...
+    auto* sender1 = testHarness.GetParticipant("Sender1", senderConfig);
+    auto* sender2 = testHarness.GetParticipant("Sender2", senderConfig);
+    auto* receiver = testHarness.GetParticipant("Receiver", receiverConfig);
+
+    auto* sender1Lifecycle = sender1->GetOrCreateLifecycleService();
+    auto* sender1TimeSync = sender1->GetOrCreateTimeSyncService();
+    auto* sender2TimeSync = sender2->GetOrCreateTimeSyncService();
+    auto* receiverTimeSync = receiver->GetOrCreateTimeSyncService();
+
+    auto* sender1Logger = sender1->Participant()->GetLogger();
```
snps-behrens: Looks like `SimParticipant` also has a `GetLogger` function. Maybe use that here instead?

SilKit/IntegrationTests/ITest_RemoteLogging.cpp
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        const auto filename = entry.path().filename().string();
+        if (filename.rfind(logNamePrefix, 0) == 0 && entry.path().extension() == ".txt")
+        {
+            candidates.emplace_back(entry.path());
+        }
+    }
+
+    if (candidates.size() != 1u)
```
snps-behrens: Would it make sense to distiguish between the two failure modes here already? `candidates.emty()` is an expected result during polling, but `candidates.size() > 1` would indicate somehting else went wrong.

snps-behrens (changes requested):
Looks pretty good already. I left some comments and questions.

SilKit/IntegrationTests/ITest_RemoteLogging.cpp
snps-behrens: I still don't understand why this polling is needed. Isn't the file created when the participant is constructed, and the `Flushlevel: Trace` ensures that the messages are flushed to file immediately? To me, the polling is turning a test failure/regression into a timeout issue with a less helpfull error message (especially with the conflation of the two error modes in the `FindLogFile` function I also metioned in the previous review).

Wouldn't a better way to do this be something like the following (I took the liberty of also using some more gtest-native string matchers):
```c++
TEST(...) {

  ...
  
  {
    SimTestHarness testHarness{testHarnessArgs};
    ...
    ASSERT_TRUE(testHarness.Run(5s));
  } 

  const auto logFiles = FindLogFiles(filePrefix);
  ASSERT_EQ(logFiles.size(), 1u) << "Expected exactly one receiver log file with prefix " << filePrefix;

  const auto logContent = ReadTextFile(logFiles.front());
  EXPECT_THAT(logContent, testing::HasSubstr(sender1Message));
  EXPECT_THAT(logContent, testing::HasSubstr(sender2Message));
}
```

SilKit/IntegrationTests/ITest_RemoteLogging.cpp
```diff
@@ -0,0 +1,191 @@
+// SPDX-FileCopyrightText: 2026 Vector Informatik GmbH
+//
+// SPDX-License-Identifier: MIT
+
+#include <system_error>
+#include <chrono>
+#include <filesystem>
+#include <fstream>
+#include <sstream>
```
snps-behrens: clangd: sstream is unused.

snps-behrens (commented):
Thanks for taking care of the things I noticed in the first review. I thought about it a bit more, and still have the question about the polling loop, but the rest looks good.

SilKit/IntegrationTests/ITest_RemoteLogging.cpp
KonradBreitsprecherBkd: Agreed, you're version is more concice. Thank you for your thorough review!

snps-behrens (approved):
Looks good to me.

PR #345, opened 2026-06-15
Fix clang-format version check

KonradBreitsprecherBkd:
Make clang-format version check more robust.


VJanKraemer (approved):
Thanks for fixing it :)

Issue #342, opened 2026-06-15
Dockerfile Warning


KamilPogorzelski2:
In .devcontainer/clangformat.Dockerfile I am getting the warning [LegacyKeyValueFromat](https://docs.docker.com/reference/build-checks/legacy-key-value-format/)

This can be easily fixed:
```Dockerfile
ENV TZ="Europe/Berlin"
ENV DEBIAN_FRONTEND="noninteractive"
ENV LC_ALL=C.UTF-8
ENV LANG=C.UTF-8
```

KonradBreitsprecherBkd:
Thanks again for pointing that out @KamilPogorzelski2 ! Will be fixed with https://github.com/vectorgrp/sil-kit/pull/347

PR #347, opened 2026-06-16
Update devcontainer dockerfile ENV format

KonradBreitsprecherBkd:
As pointed out in https://github.com/vectorgrp/sil-kit/issues/342:
Use `ENV key=value` syntax instead of depricated `ENV key value` which produces a warning.


MariusBgm (approved):
LGTM. 
So far no one of us is really using the devcontainers. maybe we should drop them in the long run?

PR #352, opened 2026-06-23
simulation: add dynamic simulation step sizes

MariusBgm:
allow modifying the simulation step size and ensure that distributed participants only advance virtual time to the minimum time step.
Changes:
- the participant config allows disabling this. by default it is off: `Experimental/TimeSynchronization/DynamicSimulationStep`
- when a remote participant is detected, and the config is either not-set or enabled, we use dynamic sim steps
  - this uses supplemental data, no protocol changes
- to fully opt out a participant has to set
```
Experimental:
    TimeSynchronization:
       DynamicSimulationStep: false
```
This is an experimental feature, ready to get user feedback.

### DIFF
--- a/SilKit/IntegrationTests/ITest_DynStepSizes.cpp
+++ b/SilKit/IntegrationTests/ITest_DynStepSizes.cpp
@@ -1,0 +1,400 @@
+// SPDX-FileCopyrightText: 2023 Vector Informatik GmbH
+//
+// SPDX-License-Identifier: MIT
+
+#include <memory>
+#include <thread>
+#include <string>
+#include <chrono>
+#include <iostream>
+#include <sstream>
+#include <functional>
+#include <vector>
+#include "ITestFixture.hpp"
+
+using namespace std::chrono_literals;
+
+namespace testing {
+namespace internal {
+template <typename Rep, typename Period>
+class UniversalPrinter<std::chrono::duration<Rep, Period>>
+{
+public:
+    static void Print(const std::chrono::duration<Rep, Period>& value, ::std::ostream* os)
+    {
+        *os << std::chrono::duration_cast<std::chrono::nanoseconds>(value).count() << "ns";
+    }
+};
+
+template <typename R1, typename P1, typename R2, typename P2>
+class UniversalPrinter<std::pair<std::chrono::duration<R1, P1>, std::chrono::duration<R2, P2>>>
+{
+public:
+    static void Print(const std::pair<std::chrono::duration<R1, P1>, std::chrono::duration<R2, P2>>& p,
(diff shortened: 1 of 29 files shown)


SilKit/IntegrationTests/SimTestHarness/SimTestHarness.cpp
```diff
@@ -304,11 +304,6 @@ void SimTestHarness::AddParticipant(const std::string& participantName, const st
     // mandatory sim task for time synced simulation
     // by default, we do no operation during simulation task, the user should override this
     auto* lifecycleService = participant->GetOrCreateLifecycleService(startConfiguration);
-    if (startConfiguration.operationMode == SilKit::Services::Orchestration::OperationMode::Coordinated)
-    {
-        auto* timeSyncService = participant->GetOrCreateTimeSyncService();
-        timeSyncService->SetSimulationStepHandler([](auto, auto) {}, 1ms);
-    }
```
KonradBreitsprecherBkd, line 311: I had a similar change when needing a harness with full control over the timeSyncService.
Removing it like this will default all harness participants to NO SYNC if they don't call `GetOrCreateTimeSyncService` in the test. I used a flag in AddParticipant(... createTimeSyncService = true) that controls the `GetOrCreateTimeSyncService` here to ensure that there is no behavior change of existing test.

SilKit/IntegrationTests/ITest_DynStepSizes.cpp
```diff
@@ -0,0 +1,360 @@
...
+            participantConfiguration += R"("DynamicSimulationStep":)";
+            participantConfiguration += participantParams.dynamicSimulationStep.value() ? "true" : "false";
+        }
+        participantConfiguration += R"(}}})";
+
+        auto&& simParticipant = _simTestHarness->GetParticipant(participantParams.name, participantConfiguration);
+        auto&& lifecycleService = simParticipant->GetOrCreateLifecycleService();
+        // The harness already created the time sync service for coordinated participants (reading the
+        // DynamicSimulationStep flag from the participant configuration above); reuse it and override the
+        // default no-op step handler below.
```
KonradBreitsprecherBkd: No longer true with the harness change

KonradBreitsprecherBkd (approved):
Only issue I found is in the SimTestHarness change. Otherwise thoroughly tested and safe.

SilKit/IntegrationTests/SimTestHarness/SimTestHarness.cpp
MariusBgm: oh wow, that's bad. I revert the whole change. i think i do no longer use it without sync in these tests...

PR #374, opened 2026-06-26
includes: do not use internal headers in public

MariusBgm:


VDanielEdwards:
`std::string` is [defined as](https://cppreference.com/cpp/string/basic_string) `std::basic_string<char>`

VDanielEdwards (approved):

SilKit/include/silkit/services/logging/string_utils.hpp
```diff
@@ -65,7 +64,10 @@ std::ostream& operator<<(std::ostream& outStream, const Level& level)
 
 inline Level from_string(const std::string& levelStr)
 {
-    auto logLevel = SilKit::Util::LowerCase(levelStr);
+    std::string logLevel = levelStr;
+    // Note: std::tolower has undefined behavior if the argument is not representable as unsigned char.
+    std::transform(logLevel.begin(), logLevel.end(), logLevel.begin(),
+                   [](unsigned char c) { return static_cast<std::string::value_type>(std::tolower(c)); });
```
VDanielEdwards:
```suggestion
                   [](char c) { return static_cast<char>(std::tolower(static_cast<int>(static_cast<unsigned char>(c)))); });
```

SilKit/source/util/StringHelpers.cpp
```diff
@@ -98,8 +98,9 @@ auto CurrentTimestampString() -> std::string
 
 auto LowerCase(std::string input) -> std::string
 {
+    // Note: std::tolower has undefined behavior if the argument is not representable as unsigned char.
     std::transform(input.begin(), input.end(), input.begin(),
-                   [](unsigned char c) { return (unsigned char)std::tolower(c); });
+                   [](unsigned char c) { return static_cast<std::string::value_type>(std::tolower(c)); });
```
VDanielEdwards:
```suggestion
                   [](unsigned char c) { return static_cast<char>(std::tolower(static_cast<int>(static_cast<unsigned char>(c)))); });
```

MariusBgm:
man i hate C. the int argument needs to be unsigned char, the return value char / std::string::value_type ... otherwise we have UB: https://en.cppreference.com/cpp/string/byte/tolower

PR #382, opened 2026-06-26
cmake: dont use internal include dirs, only the top-level

MariusBgm:
all include paths are now relative to either public silkit/ or internal SilKit/source


VDanielEdwards (approved):
LGTM! 👍 Good riddance!

PR #385, opened 2026-07-06
config: ensure yaml schema is validated

MariusBgm:
Fix checking the yaml schema in the top-level configuration. currently we only check the included yaml files against the schema validator.


KonradBreitsprecherBkd (approved):
Quickly checked, `SilKit::Config::Validate` is indeed not called (on main without includes).

PR #389, opened 2026-07-14
Update participant configuration json schema

VLukasBecker:
## Subject

- Update of participant configuration JSON schema.
  - Experimental config will be shown in the schema
  - Examples will be added to the schema
  - The schema should be equal to the schema in the schemastore repo
- Update of the configuration files 
- Validation of config files
  - git workflow which runs a python file to check the configs against the schema


SilKit/source/config/ParticipantConfiguration.schema.json
```diff
@@ -471,18 +471,6 @@
                 "type": "string",
                 "enum": [ "Simple", "Json" ]
               },
-              "DisabledTopics": {
```
KonradBreitsprecherBkd, line 474: Die gibt es doch, halt nochmal unter nem Experimental knoten siehe 
https://github.com/vectorgrp/sil-kit/commit/5cff590debdeda107c6087232049f503197a2136#diff-8f44983e4024cf000a9b498c5aefc580c452cd8feca12ee79b89c8c764d63fd4R105-R112

SilKit/source/config/ParticipantConfiguration_Full.json
```diff
@@ -258,8 +263,12 @@
     "EnableDomainSockets": false,
     "TcpSendBufferSize": 3456,
     "TcpReceiveBufferSize": 3456,
+    "AcceptorUris": [
+      "silkit://localhost:0"
```
VDanielEdwards: The `silkit://` URL schema is only used for the registry uri. The endpoints of the acceptor uris are using the `tcp://` and `local://` schemas.

```suggestion
      "tcp://0.0.0.0:0"
```

SilKit/source/config/ParticipantConfiguration_Full.yaml
```diff
@@ -185,12 +189,15 @@ Middleware:
...
+  AcceptorUris:
+  - silkit://localhost:0
   RegistryAsFallbackProxy: false
   ConnectTimeoutSeconds: 1.234
+  ExperimentalRemoteParticipantConnection: false
 Experimental:
   TimeSynchronization:
     AnimationFactor: 1.5
-    EnableMessageAggregation: Off
+    EnableMessageAggregation: 'Off'
```
VDanielEdwards, line 200: Fell down the YAML rabbit hole. Apparently, in YAML 1.1 `Off` (no quotes) parses as `false`, in YAML 1.2 it parses as the string `Off`.

.github/workflows/validate-participant-configs.yml
```diff
@@ -0,0 +1,22 @@
+name: 'Validate Participant Configuration Examples'
+on:
+  workflow_call:
```
VDanielEdwards, line 3:
```suggestion
  workflow_call:

  workflow_dispatch:
```

Just so you can easily run the workflow manually.

VDanielEdwards (approved):
Looks good, just some small things I noticed 👍

SilKit/source/config/ParticipantConfiguration.schema.json
```diff
@@ -516,22 +708,32 @@
...
+      "description": "Version of the schema used to validate this document",
+      "examples": ["1"]
+    },
+    "schemaVersion": {
+      "anyOf": [
+        { "type": "integer" },
+        { "type": "string" }
+      ],
+      "default": 0,
+      "description": "Legacy lowercase alias for SchemaVersion, kept for backwards compatibility. Prefer 'SchemaVersion'"
```
VDanielEdwards, line 721: As the only valid values (accepted by the parser) are _field not present_, `1`, or `"1"`, I propose to turn the `SchemaVersion` (and `schemaVersion`) into an `enum`:

```suggestion
    "schemaVersion": {
      "enum": [1, "1"],
      "description": "Legacy lowercase alias for SchemaVersion, kept for backwards compatibility. Prefer 'SchemaVersion'"
```

This means that the schema file describes version 1 and editors will _only_ propose that value for the field.

With `"const"` we'd have to use one of the 'alternatives' constructs (`anyOf` or `oneOf`).

VDanielEdwards (approved):
Great work! 👍 Just an idea with the `"enum"`.

SilKit/source/config/ParticipantConfiguration.schema.json
VLukasBecker: Valid idea. But I'm not sure, we would have to update the field with every version change.

PR #390, opened 2026-07-15
core: supplemental data is encoded as JSON

MariusBgm:
Supplemental data was always intended to be encoded as JSON.
This should be backwards compatible, as we use yaml parser to parse JSON and yaml.


VDanielEdwards (approved):
LGTM 👍 Tested it manually against 5.0.6 (which uses the 'old' supplemental data serialization).

PR #394, opened 2026-07-23
yaml: be more robust when parsing objects vs sequences

MariusBgm:
This makes the parsing of objects more robust.
the following was silently accepted:
```
HealthCheck:   
  - SoftResponseTimeout: 1
```
which should be invalid, but is silently accepted (because HealthCheck is a map and not a container for a sequence).

### DIFF
--- a/SilKit/source/config/YamlReader.hpp
+++ b/SilKit/source/config/YamlReader.hpp
@@ -227,13 +227,15 @@ protected:
 
     auto GetChildSafe(const std::string& name) const -> Impl
     {
-        if (HasKey(name))
+        if (IsMap())
         {
             return MakeImpl(_node.find_child(ryml::to_csubstr(name)));
         }
 
-        if (IsSequence())
+        if (_node.is_stream())
         {
+            // A leading "---" makes the document root a stream (a sequence of
+            // documents). Transparently descend into the document(s) to find the key.
             for (const auto& child : _node.cchildren())
             {
                 if (child.is_container() && HasKey(child, name))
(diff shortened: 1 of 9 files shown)


SilKit/source/config/YamlReader.hpp
```diff
@@ -227,25 +227,57 @@ class BasicYamlReader
...
     auto GetChildSafe(const std::string& name) const -> Impl
     {
-        if (HasKey(name))
+        if (IsMap())
         {
             return MakeImpl(_node.find_child(ryml::to_csubstr(name)));
         }
 
-        if (IsSequence())
+        if (_node.is_stream())
```
MariusBgm, line 235:
```suggestion
        if (IsStream())
```

SilKit/source/config/YamlReader.hpp
```diff
@@ -227,25 +227,57 @@ class BasicYamlReader
...
         {
-            for (const auto& child : _node.cchildren())
+            // A leading "---" makes the document root a stream (a sequence of
+            // documents). Transparently descend into the document(s) to find the key.
+            // A document must itself be a mapping; a document that is a bare sequence
+            // is the same object-vs-list mistype we reject below.
+            for (const auto& doc : _node.cchildren())
```
VDanielEdwards, line 239: We should only accept single-document streams, and fail if `_node.cchildren()` returns more than one document (probably also with `ThrowExpectedMapping` or create a new error message).

MariusBgm: added the check and we now fail if there are multiple documents in a stream:
```
Failed to load configuration 'stream.silkit.yaml':
SIL Kit: Configuration has errors (10):
error parsing configuration string: line 1 column 0: expected a single YAML document, but the configuration contains multiple "---"-separated documents
```

MariusBgm: i think we should disallow streams globally. the following  just chokes on the validator, not on the parser itself:
```
Description: Log to Stdout with Level Info
Logging:
---
  Sinks:
    - Level: Info
      Type: Stdout
```
gives:
```
Failed to load configuration 'stream.silkit.yaml':
SIL Kit: Configuration has errors (10):
YAML validation returned errors:
At  file : line 3 column 2: Element "Sinks" is not a valid sub-element of schema path "/"
At  file : line 4 column 6: Element "Level" is not a valid sub-element of schema path "/Sinks"
At  file : line 5 column 6: Element "Type" is not a valid sub-element of schema path "/Sinks"
```

VDanielEdwards: To be fair, the errors are accurate if both documents are processed individually. But yeah, we should also teach the validator to reject multi-document streams.

We cannot fully disallow them, since

```
---
Description: Wowee
```

is a very typical way to start YAML files (where the `---` starts the first and only document, and separates the empty list of directives from the document itself).

MariusBgm: yeah, i meant, we allow documents with a stream of exactly one document

VDanielEdwards (approved):

PR #398, opened 2026-07-29
tests: make system monitor use time sync again

MariusBgm:
Some tests which rely on the test harness' montor run flaky on systems with heavy load.
Make sure to enable timesync in the internal monitor for tests (provides back pressure and sync between participants).


KonradBreitsprecherBkd (approved):
LGTM

PR #399, opened 2026-07-29
Fix stack overflow risk in VAsioPeer::DispatchBuffer()

KamilPogorzelski2:
Replace recursive buffer dispatch with an iterative loop so multiple queued messages are processed without growing the call stack.
## Subject
Refactoring of VAsioPeer::DispatchBuffer()

## Description
When debugging an application that uses SIL Kit I was running into a stack overflow after hitting a break point. The reason for the stack overflow was a self-recursion of the VAsioPeer::DispatchBuffer() method. 

Self-recursion was replaced by a while loop without changing the behavior of the method.

### DIFF
--- a/SilKit/source/core/vasio/VAsioPeer.cpp
+++ b/SilKit/source/core/vasio/VAsioPeer.cpp
@@ -225,70 +225,71 @@ void VAsioPeer::ReadSomeAsync()
 
 void VAsioPeer::DispatchBuffer()
 {
-    if (_currentMsgSize == 0)
+    while(true) 
     {
-        if (_isShuttingDown)
+        if (_currentMsgSize == 0)
         {
-            return;
-        }
-        if (_msgBuffer.Size() >= sizeof(uint32_t))
-        {
-            std::vector<uint8_t> msgSizeInBytes(sizeof(uint32_t));
-            if (!_msgBuffer.Peek(msgSizeInBytes))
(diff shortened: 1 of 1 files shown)


KonradBreitsprecherBkd:
We had a change in our Test Harness that possibly caused sporadic test failures such as the one in the CI here. Fixed with https://github.com/vectorgrp/sil-kit/commit/51c08717f1b2b7ada4ebeef1352a828687aac4c3. Could you please rebase once more @KamilPogorzelski2?

SilKit/source/core/vasio/VAsioPeer.cpp
```diff
@@ -225,70 +225,71 @@ void VAsioPeer::ReadSomeAsync()
...
+        if (_currentMsgSize == 0 || _currentMsgSize > 1024 * 1024 * 1024)
         {
-            // not enough data to even determine the message size...
-            // restart the async read operation
-            ReadSomeAsync();
-            return;
+            _logger->MakeMessage(Services::Logging::Level::Error, TopicOf(*this))
+                .SetMessage("Received invalid Message Size: {}", _currentMsgSize.load())
+                .Dispatch();
+            Shutdown();
```
MariusBgm, line 260: we should break here.

```suggestion
            Shutdown();
            break;
```

MariusBgm (approved):
LGTM!

PR #403, opened 2026-07-30
Prevent use of fmt for log messages without arguments

KonradBreitsprecherBkd:
- Add overload for string only logs without args
- Add ITest

### DIFF
--- a/SilKit/IntegrationTests/ITest_Logging.cpp
+++ b/SilKit/IntegrationTests/ITest_Logging.cpp
@@ -1,0 +1,152 @@
+// SPDX-FileCopyrightText: 2026 Vector Informatik GmbH
+//
+// SPDX-License-Identifier: MIT
+
+#include <chrono>
+#include <filesystem>
+#include <fstream>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#include "silkit/SilKit.hpp"
+#include "silkit/services/flexray/string_utils.hpp"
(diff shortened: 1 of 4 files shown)


SilKit/IntegrationTests/ITest_Logging.cpp
```diff
@@ -0,0 +1,152 @@
...
+    return ss.str();
+}
+
+auto MakeParticipantConfiguration() -> std::string
+{
+    std::stringstream config;
+    config << R"({"Logging":{"FlushLevel":"Trace","Sinks":[)"
+           << R"({"Type":"File","Format":"Simple","Level":"Trace","LogName":")" << simpleLogName << R"("},)"
+           << R"({"Type":"File","Format":"Json","Level":"Trace","LogName":")" << jsonLogName << R"("}]}})";
+    return config.str();
```
MariusBgm:
```suggestion
auto config = R"(
Logging:
  FlushLevel: Trace
  Sinks:
    - Type: File
       Level: Trace
       Format: Simple
       LogName: ITest_logging_Simple
    - Type: File
       Level: Trace
       Format: Json
       LogName: ITest_Logging_Json
)";
```

the function can actually be a const auto config = ....

MariusBgm (approved):
just some cosmetic nits on how the configuration is built.
I really dislike JSON for tests, the yaml version is way more readable

PR #406, opened 2026-08-10
tests: Ffix ITest_AsyncSimTask repeatability by turning global into local variables

VDanielEdwards:
## Subject

Move mx, cvCounter, and cv from file-scope globals into the test body so state doesn't leak across repeated runs, and fix the wait lambdas to capture by reference instead of silently referencing globals.

- [ ] Changelog.md updated


KonradBreitsprecherBkd (approved):
LGTM! Fixes the segfault when repeating the test!

PR #407, opened 2026-08-10
Stop also in SystemState::Stopped if not already done in SystemState::Stopping

KonradBreitsprecherBkd:
A coordintated particpant ("P2") should stop as well if another participant ("P1") calls `lifecycle->Stop()`.
For P2, this is handled by monitoring the `SystemState`. If this is calculated to be `SystemState::Stopping`, then the local stop is triggered. However, this only happens if its `ParticipantState` is `Running` or `Paused`. Further, due to the local calculation of the `SystemState`, intermediate states can be skipped, in this case `SystemState::Stopping` is not guaranteed to be triggered.

This can be reproduced by spinning on `ITest_DynStepSizes - one_participant_ByOwnDuration` (Linux Debug) with `stress-ng --cpu 64`. When additionally monitoring the `SystemState` of the internalSystemMonitor in the SimTestHarness, it can be shown that  `SystemState::Running` and `SystemState::Stopping` are never triggered.

In this PR, this is duct-taped by triggering the local stop also in `SystemState::Stopped` (if it did not already happen before).

### DIFF
--- a/SilKit/source/services/orchestration/LifecycleService.hpp
+++ b/SilKit/source/services/orchestration/LifecycleService.hpp
@@ -178,6 +178,9 @@ private:
     // for immediate checks (e.g., not to send out the NextSimTask after a stopping in the SimTaskHandler).
     std::atomic<bool> _stopRequested{false};
     std::atomic<bool> _pauseRequested{false};
+
+    bool _stopRequestedDueToSystemStateChange{false};
+
 };
 
 // ================================================================================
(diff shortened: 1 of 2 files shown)


SilKit/source/services/orchestration/LifecycleService.hpp
```diff
@@ -178,6 +178,9 @@ class LifecycleService
     // for immediate checks (e.g., not to send out the NextSimTask after a stopping in the SimTaskHandler).
     std::atomic<bool> _stopRequested{false};
     std::atomic<bool> _pauseRequested{false};
+
+    bool _stopRequestedDueToSystemStateChange{false};
```
VDanielEdwards, line 182: If TSAN / UBSAN isn't hitting anything this can stay as a `bool`, I'm just seeing that `_stopRequested` is an `std::atomic<bool>`.

VDanielEdwards (approved):
As far as I can tell this change should not create more issues, and solves one.

PR #416, opened 2026-08-17
log: fix missing arguments in timesync

MariusBgm:
fix log messages with missing argument in timesync service.


KonradBreitsprecherBkd (approved):
Looks good.

PR #419, opened 2026-09-01
dashboard: get rid of oatpp

MariusBgm:
## Drop oat++ from the SIL Kit Dashboard client

- **removed:** `ThirdParty/oatpp` submodule, its CMake and license entries
- **added:** minimal HTTP/1.1 client over the bundled standalone asio (`dashboard/http/`)
- **changed:** dashboard JSON now emitted with the bundled rapidyaml
- **changed:** wire format identical bar 3 cosmetics: space after `:`, no `\/`, UTF-8 over `\uXXXX`; unescapable control chars become U+FFFD
- **fixed:** registry hung on shutdown when the dashboard accepted but never answered; requests now have connect/write/read deadlines and abort after a grace period
- **fixed:** use-after-free, `RegistryInstance` destroyed the dashboard before the registry pointing at it
- **fixed:** one unmappable event killed the worker thread, silently ending all dashboard reporting; failures are now logged and skipped per event and per flush
- **fixed:** unhandled controller types threw instead of being skipped, reachable by 11 of 21 types
- **removed:** bulk-update probe, which killed the worker on a negative result; also drops a startup round-trip and a ~7 s stall when unreachable
- **fixed:** events pending at shutdown were discarded, now flushed
- **fixed:** dashboard log lines had no `Dashboard` topic and escaped topic filtering
- **changed:** `SILKIT_BUILD_DASHBOARD=OFF` with `--dashboard-uri` now reports the missing feature plainly
- **added:** warning when the dashboard is on but `CollectFromRemote` is off, which silently yields no metrics or attributes
- **changed:** `SilKitEvent` type erasure to `std::variant`, worker split out; net −348 lines
- **tests:** new suites for HTTP client, parser, retry, JSON writer, DTO mapper, worker, shutdown; `IRestClient` mocked, batching had no coverage
- **tests:** no sleeps or wall-clock assertions, latch handover, stable under CI load


SilKit/source/dashboard/http/AsioHttpClient.cpp
```diff
@@ -0,0 +1,523 @@
...
+{
+    SilKit::Services::Logging::ILoggerInternal* logger;
+    std::string host;
+    uint16_t port;
+    std::string hostHeader;
+    AsioHttpClientTimeouts timeouts;
+
+    asio::io_context ioContext{1};
+    std::optional<asio::ip::tcp::socket> socket;
+    asio::streambuf readBuffer;
```
VDanielEdwards, line 50: Default `max_size` is `SIZE_MAX`, so `ReadHead`'s `async_read_until` and `ReadUntilClose`'s `async_read` never self-terminate — `maxHeadSize`/`maxHttpBodySize` are only checked once the data is already buffered. A server that streams headers without the blank line, or a response with neither `Content-Length` nor `Transfer-Encoding`, grows RSS for the whole 30s read window.

`asio::streambuf readBuffer{maxHeadSize + maxHttpBodySize};` bounds both (`async_read_until` then completes with `error::not_found`). Content-Length and chunked framing are already bounded before reading.

SilKit/source/dashboard/http/AsioHttpClient.cpp
```diff
@@ -0,0 +1,523 @@
...
+    /*! Cancel the pending operation and let its handler run.
+     *
+     *  Without draining, the handler would later write through a dangling reference to the caller's
+     *  stack-allocated error_code.
+     */
+    auto DrainAndFail(std::error_code reason) -> std::error_code
+    {
+        DropSocket();
+        ioContext.restart();
+        ioContext.run();
```
VDanielEdwards: Unbounded on the resolve path: asio's resolver keeps io_context work outstanding until its background `getaddrinfo` returns, and nothing cancels that mid-call, so `run()` blocks for the full OS resolver timeout (~40s glibc default). The connect deadline detects the DNS timeout and then waits it out, and the `Abort()` post can't end it either — so "every step has a deadline" doesn't hold for DNS. Only affects hostname URIs, not IPs.

SilKit/source/dashboard/service/DashboardDtoMapper.cpp
```diff
@@ -35,7 +38,7 @@ auto ToUInt64(const std::string& value) -> std::uint64_t
     }
 }
```
VDanielEdwards: No return on this path — both `if`s are runtime, not `if constexpr`. Pre-existing, but cheap to fix while the file moves.

SilKit/source/dashboard/http/AsioHttpClient.cpp
```diff
@@ -0,0 +1,523 @@
...
+        }, timeouts.read);
+        if (ec)
+        {
+            return ec;
+        }
+        if (headSize > maxHeadSize)
+        {
+            return asio::error::message_size;
+        }
+        const auto* data = asio::buffer_cast<const char*>(readBuffer.data());
```
VDanielEdwards, line 251: `asio::buffer_cast` is deprecated in asio 1.30 — compiles only while `ASIO_NO_DEPRECATED` is unset. `static_cast<const char*>(readBuffer.data().data())` is the current spelling. Also lines 277, 298, 368 and in `FakeHttpServer.hpp`.

SilKit/source/dashboard/json/DashboardJsonWriter.hpp
```diff
@@ -0,0 +1,128 @@
...
+    {
+        WriteQuoted(value);
+    }
+
+    // Guard the paths that would silently emit an unquoted scalar.
+    void Write(const char*) = delete;
+    void Write(bool) = delete;
+
+    //! Matches oatpp's OATPP_FLOAT_STRING_FORMAT, so statistic values keep their previous digits.
+    void Write(double value)
```
VDanielEdwards: `%.16g` emits bare `nan`/`inf` for a non-finite statistic, which is invalid JSON. oatpp did the same, so not a regression.

SilKit/source/dashboard/http/Test_AsioHttpClient.cpp
```diff
@@ -0,0 +1,231 @@
...
+    AsioHttpClient client{nullptr, "127.0.0.1", server.Port()};
+
+    EXPECT_FALSE(client.Post("a", "{}").transportError);
+    EXPECT_FALSE(client.Post("b", "{}").transportError);
+
+    EXPECT_EQ(server.AcceptCount(), 1) << "the keep-alive connection should be reused";
+    EXPECT_EQ(server.Requests().size(), 2u);
+}
+
+TEST_F(Test_AsioHttpClient, Post_ReconnectsWhenTheServerClosesTheConnection)
```
VDanielEdwards: This covers the `Connection: close` path. The stale-keep-alive second attempt in `Post` (server closes an idle connection without saying so) isn't exercised — `FakeHttpServer{AlwaysReply(Reply(200, "{}")), /*closeAfterReply=*/true}` would hit it.

SilKit/source/dashboard/http/Test_AsioHttpClient.cpp
```diff
@@ -0,0 +1,231 @@
...
+{
+    FakeHttpServer server{AlwaysReply("this is not an HTTP response\r\n\r\n")};
+    AsioHttpClient client{nullptr, "127.0.0.1", server.Port(), FastTimeouts()};
+
+    EXPECT_TRUE(client.Post("a", "{}").transportError);
+}
+
+TEST_F(Test_AsioHttpClient, Post_ReportsATransportErrorWhenNothingIsListening)
+{
+    // Port 1 is reserved and never has a listener. The connect deadline is out of reach, so a
```
VDanielEdwards, line 159: 1h connect deadline + port 1 assumes the host *refuses* rather than drops. Where loopback port 1 is filtered this hangs to the CTest timeout. Same in `service/Test_DashboardShutdown.cpp:79`.

SilKit/source/dashboard/DashboardInstance.hpp
```diff
@@ -4,39 +4,42 @@
...
     , private SilKit::Core::IRegistryEventListener
 {
-    struct SimulationData
-    {
-        SystemStateTracker systemStateTracker;
-    };
-
 public:
-    explicit DashboardInstance();
+    //! Throws if dashboardUri is not a usable http URI, so a bad --dashboard-uri fails at creation.
```
VDanielEdwards: `Uri::Parse` only requires a scheme and a non-empty host — `silkit://foo` passes, and a missing port silently becomes 8500 rather than 80. Comment overstates the validation.

SilKit/source/dashboard/DashboardInstance.cpp
```diff
@@ -37,271 +48,64 @@ using VSilKit::ServiceData;
...
-        else
-        {
-            _logger->MakeMessage(Level::Debug, TopicOf(*this))
-                .SetMessage("Dashboard bulk-updates are not available, falling back to individual requests")
-                .Dispatch();
-        }
+    // Parse eagerly so that a malformed --dashboard-uri is reported when the instance is created,
+    // rather than later on the registry's thread. The result is discarded; DashboardRestClient
+    // parses it again once it is built.
+    (void)SilKit::Core::Uri::Parse(dashboardUri);
```
VDanielEdwards, line 61: Consider making sure that the URI scheme is `http`. The `Uri::Parse` function accepts any scheme.

```suggestion
    const auto uri = SilKit::Core::Uri::Parse(dashboardUri);
    if (uri.Scheme() != "http")
    {
        SilKit::SilKitError{"Dashboard URI must have the http scheme!"};
    }
```

Alternatively the check could be done in `DashboardRestClient`.

SilKit/source/dashboard/CMakeLists.txt
```diff
@@ -61,49 +76,51 @@ if(SILKIT_BUILD_DASHBOARD)
...
+
+    foreach(testSource IN LISTS SILKIT_DASHBOARD_TEST_SOURCES)
+        add_silkit_test_to_executable(SilKitDashboardTests
+            SOURCES ${testSource}
+            LIBS
+            S_SilKitImpl
+            O_SilKit_Dashboard
+            I_SilKit
+        )
+    endforeach()
```
VDanielEdwards, line 123: Using `target_sources` is only valid when `SILKIT_BUILD_TESTS=ON`. Otherwise the target does not exist. All sources can be combined into a single call to `add_silkit_test_to_executable`. This does the check internally.

```suggestion
    add_silkit_test_to_executable(SilKitDashboardTests
        SOURCES
        http/FakeHttpServer.hpp
        Mocks/MockRestClient.hpp
        Test_DashboardEventQueueWorker.cpp
        Test_DashboardSilKitEventQueue.cpp
        client/Test_DashboardSystemServiceClient.cpp
        http/Test_AsioHttpClient.cpp
        http/Test_HttpResponseParser.cpp
        http/Test_RetryingHttpClient.cpp
        json/Test_DashboardJsonWriter.cpp
        service/Test_DashboardDtoMapper.cpp
        service/Test_DashboardRestClient.cpp
        service/Test_DashboardShutdown.cpp
        LIBS
        S_SilKitImpl
        O_SilKit_Dashboard
        I_SilKit
    )
```

SilKit/source/dashboard/json/DashboardJsonWriter.hpp
```diff
@@ -0,0 +1,128 @@
...
+    {
+        char buffer[64];
+        const int length = std::snprintf(buffer, sizeof buffer, "%.16g", value);
+        node << ryml::csubstr{buffer, static_cast<size_t>(length < 0 ? 0 : length)};
```
VDanielEdwards, line 57: Since `snprintf(..., "%.16g", ...)` can emit `nan` or `inf` (for which JSON has no represantation in any case) we should handle this case here. I'd just emit 0. We could also add a log message in this cae, but then we'd have to keep a logger pointer around.

```suggestion
        if (std::isfinite(value))
        {
            char buffer[64];
            const int length = std::snprintf(buffer, sizeof buffer, "%.16g", value);
            node << ryml::csubstr{buffer, static_cast<size_t>(length < 0 ? 0 : length)};
        }
        else
        {
            node << 0;
        }
```

SilKit/source/dashboard/http/IHttpClient.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+    //! Response body; only populated when transportError is false.
+    std::string body;
+};
+
+/*! A minimal blocking HTTP client, sufficient for the dashboard's three POST endpoints.
+ *
+ *  Implementations must never throw out of Post(): every failure is reported as
+ *  HttpResult::transportError. Post() is expected to be called from a single thread; Abort() may be
+ *  called concurrently from another.
```
VDanielEdwards, line 31: Can `Reset()` also only be called from the same, singular thread as `Post(...)`?

SilKit/source/dashboard/http/AsioHttpClient.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include "dashboard/http/HttpResponseParser.hpp"
+#include "dashboard/http/IHttpClient.hpp"
+
+#include "services/logging/ILoggerInternal.hpp"
+
+// asio types are kept out of this header so that the 22k-line asio headers are confined to the
+// single translation unit that implements the client.
+namespace asio {
+class io_context;
```
VDanielEdwards, line 23: Is this forward declare neccessary? AFAICT its not used in this header, and the implementation includes the asio headers themselves.

VDanielEdwards (changes requested):
Heres a mixed set of comments made by Claude and me.

SilKit/source/dashboard/http/AsioHttpClient.cpp
VDanielEdwards: We could set `ASIO_NO_DEPRECATED` for the SIL Kit build.

SilKit/source/config/BasicYamlWriter.hpp
```diff
@@ -0,0 +1,130 @@
+#pragma once
```
VDanielEdwards, line 1: General formatting question: would you prefer `#pragma once` above or below the SPDX header? Personally I prefer it to be below the header. Compare with `SilKit/source/core/vasio/IRegistryEventListener.hpp`.

Not relevant for the approval though.

VDanielEdwards (approved):
👍 Looks very nice 😻 I've also let Claude review the changes, focusing on the networking and ser/des code and it didn't bring up anything significant that held up to further scrutiny.

SilKit/source/config/BasicYamlWriter.hpp
MariusBgm: my gut reaction would be to have `#pragma once` at the top, gives me an immediate feedback that this is a header, without looking at the vim status bar / filename. but to be honest, i don't truly care. i'm in the go-fmt camp: just give me a single default formatting standard to use, and i will hapily adhere to the formatting standard :-). 

spdx + pragma is fine, i will make a PR that adjusts the formatting in the whole project.

Issue #420, opened 2026-09-04
Bug: PcapSink::Trace() is not actually thread-safe, pcap output is corrupted under concurrent tracing


nourfakhfakh93:
<img width="670" height="210" alt="Image" src="https://github.com/user-attachments/assets/77dfbcd7-51f1-4205-9b1f-a37c21547634" />

std::unique_lock<decltype(_lock)> lock;

This default-constructs a unique_lock that is not associated with _lock, it never actually locks the mutex. It looks like the intent was std::unique_lock<decltype(_lock)> lock(_lock);.

Waiting for your confirmation whether this is a bug, and waiting for the release containing the fix if so.

Thank you in advance.

MariusBgm:
thanks for reporting.that is indeed a bug. that will bite you if you have transmits in parallel to receives. 
Also, did you know that you can just easily link to code in the repo, without taking a screenshot. just click on the line number, and select "Copy permalink", and paste that line like this: https://github.com/vectorgrp/sil-kit/blob/75fd429ad3fb702998d5771fd2aaba6300cb7227/SilKit/source/tracing/PcapSink.cpp#L110

PR #421, opened 2026-09-07
cleanup: remove dead mutexes, properly take lock

MariusBgm:
Sweep across the source tree, check what mutexes are not properly taken or have not been in use at all.
Triggered by #420


VDanielEdwards:
Yikes! Good catches 💯

PR #423, opened 2026-09-11
core: replace SharedVector with SharedSpan to remove payload copies

MariusBgm:
SharedVector could only own a whole vector and view all of it. SharedSpan separates what keeps the bytes alive from what is visible:

    std::shared_ptr<const void> _owner;   // may be null
    Span<const T> _view;

shared_ptr<U> converts to shared_ptr<const void> while keeping U's deleter in the control block, so the owner can be type erased and a payload can alias a slice of a larger buffer. One type now covers an owning copy, a view into a buffer owned elsewhere, and a borrowed view.

Send path:

The copy of the user's payload at the wire conversion stays and is the only one. It is mandatory: VAsioConnection::SendMsg defers to the I/O thread via asio::post, so the caller's Span is already dangling by the time serialization runs. Everything after that copy is refcounted.

VAsioTransmitter no longer re-serializes once per remote peer. The serialized form differs per receiver only in the remoteIndex field of the network header, so the message is serialized once; each peer copies the 29 byte header, patches its own remoteIndex, and writes header plus shared body as a two buffer gather write. The send queue holds the body by reference count, so it outlives the slowest peer, including across a disconnect mid-write.

Receive path:

The copy out of the ring buffer stays, since the ring wraps and a message may be physically discontiguous, but it now lands in a shared_ptr owned blob, one per message. Deserialized byte payloads alias that blob via a range checked Subspan instead of being copied out of it again. The blob is anchored by the MsgT local in VAsioReceiver::ReceiveRawMsg for the whole handler dispatch.

Also:

- ExecuteOnIoThread captured its arguments by copy, silently discarding the move that the rvalue SendMsg overloads perform. It now moves them.
- MessageBuffer no longer zero fills bytes it immediately overwrites, and the stale static SerializedSize hint is replaced by a per message estimate.
- The per received message 4 byte heap peek is now a stack array, which also fixes a misaligned read through reinterpret_cast.
- RemoteServiceEndpoint references the descriptor instead of deep copying it.
- RPC wire types move off plain std::vector, dropping the ToStdVector copies.

Public API, ABI and wire format are unchanged; SharedSpan does not appear in any installed header and no protocol version bump is needed. The Ethernet runt padding buffer is deliberately left as is: it is correct under the documented callback contract, and changing it only matters for a retained payload API that is not being added.

Received payloads must not reach MessageHistory, which retains for the participant's lifetime. That holds structurally today, since VAsioTransmitter is only reachable from DistributeLocalSilKitMessage; it is documented at both ends with SharedSpan::Cloned() as the escape hatch.


SilKit/source/util/SharedSpan.hpp
VDanielEdwards, line 1: This is really neat ❤️ Something like this might be handy for certain strings too.

SilKit/source/core/vasio/VAsioPeer.hpp
```diff
@@ -85,12 +88,30 @@ class VAsioPeer
...
     // Private Methods
+    /*! \brief One queued write: an optional per-peer header followed by a shared body.
+     *
+     * A message that is sent to several peers differs only in the remote index inside its network
+     * header, so the body can be shared between peers and only the small header is per-peer. The
+     * header is stored inline; headerSize == 0 marks an item that consists of the body alone.
+     */
+    struct SendItem
+    {
+        static constexpr size_t kMaxHeaderSize = 32;
```
VDanielEdwards, line 99: Google-style constant name.

SilKit/source/core/vasio/VAsioConnection.hpp
```diff
@@ -444,7 +444,13 @@ class VAsioConnection
     template <typename... MethodArgs, typename... Args>
     inline void ExecuteOnIoThread(void (VAsioConnection::*method)(MethodArgs...), Args&&... args)
     {
-        _ioContext->Post([=]() mutable { (this->*method)(std::move(args)...); });
+        // NB: the arguments are stored in a tuple by-value so that rvalue arguments are moved rather
+        //     than copied. A plain [=] capture would copy them, discarding the move that the rvalue
+        //     SendMsg overloads deliberately perform. The closure must stay copy-constructible
+        //     because IIoContext::Post takes a std::function.
+        _ioContext->Post([this, method, args = std::make_tuple(std::forward<Args>(args)...)]() mutable {
```
VDanielEdwards, line 451: I'd love to switch `Post` to `std::move_only_function` but that is C++23 unfortunately. Then we could only move here and avoid copies of the callable entirely.

SilKit/source/core/vasio/SerializedSizeHint.hpp
```diff
@@ -0,0 +1,63 @@
...
+ * a field written after a large payload can trigger a reallocation that copies the payload.
+ *
+ * The hint may over- but must never systematically under-estimate for payload carrying types.
+ * It is only a capacity hint, so an inaccurate value costs performance, never correctness.
+ */
+template <typename MessageT>
+struct SerializedSizeHint
+{
+    //! Network headers plus room for the fixed size fields of small messages.
+    static constexpr size_t kFixedOverhead = 128;
```
VDanielEdwards, line 31: Google-style constant. Also below in the macro used to generate specializations.

Demos/tools/Benchmark/PerformanceTests.py
```diff
@@ -286,28 +291,47 @@ def assess_test(test: Test, reference: ConfigRepository, under_test: ConfigRepos
...
+    if regressions:
+        print(f"Regressions detected in: {', '.join(regressions)}")
+
+    return not regressions
+
 
 ##### start script #####
 
-DEFAULT_REFERENCE_VERSION = "v4.0.52"
+DEFAULT_REFERENCE_VERSION = "v5.0.7"
```
VDanielEdwards, line 334: I think the idea was to compare to an older version of SIL Kit (and have the reference version configurable). Does it make sense to raise it to the most recent release (as of today)?

MariusBgm: i am refactoring the script anyways, it defaulted to msbuild. will bump it to use ninja if available and fix the fallout for oatpp cmake

Demos/tools/Benchmark/performance-tests.toml
```diff
@@ -1,5 +1,7 @@
 [repositories.reference]
-version = "v4.0.52"
+# NB: compare against the immediately preceding release. A far older reference folds years of
```
VDanielEdwards, line 2: I partially disagree with this comment. It makes sense to run a performance analysis against the preceding release (to see if changes in a PR made thing better or worse and how much), but if we want to compare different changes (that might not be rebased to the latest release e.g.), we need a stable, common reference version.

Demos/tools/Benchmark/PerformanceTests.py
VDanielEdwards: Regarding `oatpp`: adding the `CMAKE_POLICY_VERSION_MINIMUM=3.9` or whatever doesn't hurt the build. Especially if it is specific for benchmarking, so we can just keep that in. It should just not be present in 'normal' builds (i.e., the presets).

SilKit/source/core/vasio/SerializedSizeHint.hpp
MariusBgm: argh right. I thought we used no decorations, like `static constexpr size_t FixedOverhead`. but i also found some screaming snake case, e.g. `FIXED_OVERHEAD`. do we have a style preference for that?

VDanielEdwards: Personally I like screaming snake case for constant (`constexpr` / ...) values. I just checked the core guidelines, and they recommend using screaming snake case only for macros. So I'll refrain from using those in the future. `PascalCase` for public values, `camelCase` for private ones seems sensible to me.

MariusBgm:
the performance improvements for big messages / many receivers look good. for a single remote receiver there is a slight regression, because we have an extra allocation. TODO adding a fast path for single peer, and inlining small messages not just headers

MariusBgm:
@VDanielEdwards  i changed the benchmark tool to provide output to my liking (and reusing build dir), we can easily revert those independent of the improvements in the code here.
One can run the benchmark using `uv run ./Demos/tools/Benchmark/PerformanceTests.py --under-test-bin-dir _build/release/Release`

Did some benchmarks, on my (very loaded) windows11 workstation  and WSL (i9-13950HX, 128GB):
```
Test                                         Unit          Reference        Under test  Change  Verdict
-------------------------------------------  -----  ----------------  ----------------  ------  --------
Latency                                      us     14.7971 +/-73.2%  12.3404 +/-60.8%  -16.6%  PASSED
Throughput (large messages)                  MiB/s   2377.43 +/-3.6%   3507.05 +/-5.7%  +47.5%  IMPROVED
Throughput (small messages)                  MiB/s   3.46634 +/-1.8%   3.3422 +/-19.3%   -3.6%  FAILED
Throughput (large messages, fan-out of 8)    MiB/s   4551.79 +/-2.6%   8474.56 +/-9.0%  +86.2%  IMPROVED
Message rate (small messages, fan-out of 8)  1/s      732086 +/-2.9%    674514 +/-2.7%   -7.9%  FAILED
```
WSL on same machine:
```
Test                                         Unit              Reference          Under test   Change  Verdict
-------------------------------------------  -----  --------------------  ------------------  -------  --------
Latency                                      us         17.5604 +/-57.0%     18.195 +/-52.8%    +3.6%  PASSED
Throughput (large messages)                  MiB/s       3356.09 +/-6.3%    4992.45 +/-13.0%   +48.8%  IMPROVED
Throughput (small messages)                  MiB/s       7.97024 +/-1.6%     7.53976 +/-3.6%    -5.4%  FAILED
Throughput (large messages, fan-out of 8)    MiB/s       6563.92 +/-4.7%     13696.4 +/-6.3%  +108.7%  IMPROVED
Message rate (small messages, fan-out of 8)  1/s    1.25645e+06 +/-22.5%  1.3211e+06 +/-4.6%    +5.1%  PASSED
```
on a bare metal Ubuntu 24.04.4 (i9-13900K, 64GB):
```
Test                                         Unit             Reference           Under test   Change  Verdict
-------------------------------------------  -----  -------------------  -------------------  -------  --------
Latency                                      us        3.20371 +/-18.7%     3.06338 +/-17.3%    -4.4%  PASSED
Throughput (large messages)                  MiB/s      4765.33 +/-4.8%      7265.37 +/-9.4%   +52.5%  IMPROVED
Throughput (small messages)                  MiB/s       10.665 +/-0.9%      10.6008 +/-1.2%    -0.6%  PASSED
Throughput (large messages, fan-out of 8)    MiB/s      13661.7 +/-6.2%      34721.1 +/-7.1%  +154.1%  IMPROVED
Message rate (small messages, fan-out of 8)  1/s    4.02319e+06 +/-6.0%  4.45354e+06 +/-5.9%   +10.7%  PASSED
```

tweaking the inline-message-size from 128 to 256 gives some improvements on WSL, but nothing justifiable.


==> i think this is still valuable. the high variance of test results on my machine may be just noise.

SilKit/source/util/SharedSpan.hpp
```diff
@@ -0,0 +1,204 @@
...
+    /*! \brief View elements whose storage is kept alive by owner.
+     *
+     * Precondition: view refers to elements inside the object owned by owner. Prefer Subspan() or
+     * MakeSharedSpan() over calling this directly, as those range-check.
+     */
+    template <typename U>
+    SharedSpan(std::shared_ptr<U> owner, Span<const T> view);
+
+    //! \brief View elements without taking ownership. The caller guarantees their lifetime.
+    static auto Borrowed(Span<const T> view) -> SharedSpan;
```
VDanielEdwards: Yep, @snps-fricke was right, `Borrowed` isn't used outside of the test code. I checked in the `dev/zero_copy_opt` as well.

